PR #4, opened 2022-09-13
feat: Auto dark mode

efstajas:
Automatically enables the Radicle Design System's dark mode depending on system-level user preference.

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -1,12 +1,33 @@
 <script>
+  import { onMount } from 'svelte';
+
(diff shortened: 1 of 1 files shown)


evvvritt:
Maybe let's hold off for now until we see initial branding?

efstajas:
@evvvritt IMO regardless of what the branding looks like, we'd want a dark / light theme — we should of course customize the colors of the theme according to the brand by overriding the vars set in `colors.css`

evvvritt (approved):

PR #6, opened 2022-09-15
feat: force connection to default network if network unsupported

efstajas:
Checks if the current network is supported, and automatically triggers the wallet's network switch functionality if not.


efstajas:
@jtourkos I updated this to use the `SUPPORTED_CHAINS` constant from the Drips SDK, please have a look :)

src/lib/stores/wallet/wallet.store.ts
```diff
@@ -74,6 +79,10 @@ export default (() => {
     _clear();
   }
 
+  function _networkIsSupported(network: Network): boolean {
```
jtourkos, line 82: Extremely minor (and personal preference 🤓): 

Consider naming boolean methods _starting_ with the "logical" prefix, like `isNetworkSupported`, 'hasDoneThat`, etc.

src/lib/stores/wallet/wallet.store.ts
```diff
@@ -59,9 +59,14 @@ export default (() => {
     if (!browser) throw new Error('Can only connect client-side');
 
     const instance = await web3Modal.connect();
-
     const provider = new ethers.providers.Web3Provider(instance);
 
+    if (!_networkIsSupported(await provider.getNetwork())) {
```
jtourkos, line 64: Keep in mind that if you try to do `AddressAppClient.create(unsupportedProvider)` it will throw an `unsupportedNetwork` `DripsError`. Just saying this for the "drips store".

jtourkos (approved):

PR #7, opened 2022-09-15
feat: ens store

efstajas:
Adds a new `ens` store that provides a cache-enabled `lookup(address)` function. After calling this function and an ENS identity is discovered, results can be found at `$ens[address]`.


src/lib/components/connect-button/connect-button.svelte
```diff
@@ -1,10 +1,31 @@
...
+      // Connect all stores to wallet
+      ens.connect(provider);
+
+      // Lookup own name
+      ens.lookup(address);
+    } else {
+      // Disconnect all stores from wallet
+      ens.disconnect();
+    }
+  }
```
efstajas, line 24: Not sure if this work of connecting & disconnecting stores that require a wallet here in the scope of the login button is necessarily the best place, as we may have other entrypoints for connecting and it might somehow happen that this reactive statement only starts listening after the store is initialized. Maybe moving this to a separate module somewhere might make more sense?

src/lib/components/connect-button/connect-button.svelte
```diff
@@ -1,10 +1,31 @@
 <script lang="ts">
   import Button from 'radicle-design-system/Button.svelte';
-  import walletStore from '$lib/stores/wallet';
+  import wallet from '$lib/stores/wallet';
+  import ens from '$lib/stores/ens';
+
+  async function connect() {
```
jtourkos, line 6: Question for the future: do we prefer "functions" or const myFunc = () => {} format?

src/lib/components/connect-button/connect-button.svelte
efstajas: >  it might somehow happen that this reactive statement only starts listening after the store is initialized

Not a concern after all as the statement will evaluate once on initialisation of the component, even without a change in any of the tracked variables.

jtourkos (approved):

src/lib/components/connect-button/connect-button.svelte
efstajas: I don't have a strong preference either way, do you?

jtourkos: No, no strong opinion. I am used to `const`, but let's keep it as is; no reason to change.
To know so that we are consistent.

PR #8, opened 2022-09-20
feat: drips store

efstajas:
Adds a new Drips store that creates instances of all Drips SDK clients and exposes them in its state.

In the future, we may add more drips-related information to the state of this store (such as info about the current cycle), which is why using a store pattern to share these client instances globally makes sense.

### DIFF
--- a/src/lib/stores/drips/drips.store.ts
+++ b/src/lib/stores/drips/drips.store.ts
@@ -1,0 +1,42 @@
+import type { Web3Provider } from '@ethersproject/providers';
+import { AddressAppClient, DripsSubgraphClient } from 'radicle-drips';
+import { writable } from 'svelte/store';
+
+interface State {
+  clients: {
+    addressApp: AddressAppClient;
+    subgraph: DripsSubgraphClient;
+  };
+}
+
+export default (() => {
+  const state = writable<State | undefined>();
+
+  /**
+   * Connect the store to a provider and initialize all Drips clients at `state.clients`.
+   * @param toProvider The provider to connect to.
+   */
+  async function connect(toProvider: Web3Provider): Promise<void> {
+    const { chainId } = toProvider.network;
+
+    state.set({
+      clients: {
+        subgraph: DripsSubgraphClient.create(chainId),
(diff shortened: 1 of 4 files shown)


src/lib/stores/drips/drips.store.ts
```diff
@@ -0,0 +1,44 @@
+import type { Web3Provider } from '@ethersproject/providers';
+import { AddressAppClient, DripsHubClient, DripsSubgraphClient } from 'radicle-drips';
+import { writable } from 'svelte/store';
+
+interface State {
+  clients: {
+    addressApp: AddressAppClient;
+    hub: DripsHubClient;
```
jtourkos, line 8: Minor: to keep it consistent with the others, I'd name it `dripsHub`.

src/lib/stores/drips/drips.store.ts
```diff
@@ -0,0 +1,44 @@
...
+   * @param toProvider The provider to connect to.
+   */
+  async function connect(toProvider: Web3Provider): Promise<void> {
+    const { chainId } = toProvider.network;
+
+    state.set({
+      clients: {
+        subgraph: DripsSubgraphClient.create(chainId),
+        addressApp: await AddressAppClient.create(toProvider),
+        hub: await DripsHubClient.create(toProvider),
```
jtourkos, line 27: One small point on the SDK design. There are 2 ways to have "access" to a `DripsHubClient` instance.

1. As you do it here, by creating a new instance passing in a `provider`.
2. From an `AddressAppClient` instance itself. The `AddressAppClient.dripsHub` property returns a DripsHubClient connected to the same provider as the `AddressAppClient.

If we only create client instances here, it doesn't matter.  Just FYI.

(You might not even need a `DripsHubClient`. The only "interesting" method at the moment is "squeezing" - for now)

jtourkos (approved):

src/lib/stores/drips/drips.store.ts
efstajas: 👍 I see. Makes sense! Gonna remove the dripsHub instance for now, then.

PR #9, opened 2022-09-22
feat: Add Button & List Select representational components

efstajas:


efstajas:
@evvvritt 

> I think our current Figma buttons use "small plus"? 🤓

Ya, the icon is just a prop, for the showcase I just used some random icon. In practice you can pass any component as `icon`:) 

> I personally prefer if multi-select with a circle next to the logo that get's filled on selection (conveys it's a multi-select before selecting)

Ya, good point, I will change that

> filter/multi-select - nice if up/down arrows move you through the rows (not required though)

It already fully supports tab navigation (tab to go down, shift-tab to go up) and selection with enter as minimum accessibility requirements. I'll look into adding arrow key support too 👍

efstajas:
@evvvritt did done, please have a look!

efstajas:
Btw, just in case y'all notice this while looking in the code… I'm using a `hidden` class instead of a simple conditional render `{#if}` statement for hiding filtered items in the list because with a conditional render, they actually get removed from the DOM which causes the image to flicker when it gets added back. With the `hidden` class, this doesn't happen.

evvvritt (approved):
looks good 👍

PR #10, opened 2022-09-23
feat: header, identity badge, scroll store, account menu skeleton

efstajas:
Adds:

- basic sticky header component as part of the base layout, plus a global page container with appropriate margins
Screenshot 2022-09-23 at 10 49 00

- A `Flyout` component that allows easily building dropdown interactions such as the account menu, together with a placeholder for the account menu dropdown.

Screenshot 2022-09-23 at 10 49 31

- A `scroll` store which allows globally reading the current scroll position from a single listener, in addition to locking & unlocking the root container's scrolling (which we will need for example to prevent scrolling of the background while a modal is open).

- New `IdentityBadge` component that takes an address and automatically works with the ENS store to resolve any ens name, and then fade to that if found. If not, it displays a collapsed address & blockie avatar placeholder.

Screenshot 2022-09-23 at 10 52 00


efstajas:
Still writing tests

src/routes/+layout.svelte
```diff
@@ -22,10 +24,15 @@
   });
 
   onMount(() => walletStore.initialize());
+
+  onMount(() => scroll.attach());
```
jtourkos, line 28: Don't know if this is the "Svelte way", but why have 2 onMount implementations?

jtourkos (approved):

src/routes/+layout.svelte
efstajas: Good question, not necessarily aware of this being "the svelte way" though I've seen it in a few projects.

My thinking was to separate them as distinct mount calls given they're separate concerns, and sequence doesn't matter, which would be a tad less explicit in one mount function. 🤷‍♂️

I can also see why it might look awkward though, if you think it'd be better I'm happy to change it 🙂

jtourkos: No no, it's okay like this. I just wanted to know if there was a convention I didn't know about.
Let's do it this way (and just keep the `onMount` implementations "close enough" in the source code so that we can quickly locate them if they are more than one.)

PR #12, opened 2022-09-26
feat: tokens store

### LINKED ISSUE #13: Tokens store
A store that:

- exposes a default token list, multi-chain
- serves token icon URLs
- Allows adding and deleting custom tokens
- Stores and serves custom tokens from localstorage
- Allows banning and unbanning custom tokens

efstajas:
Adds a new store that serves Uniswap's default token list for a given chain, and allows adding, removing and banning/unbanning custom tokens, which are persisted in localstorage.

Closes #13

### DIFF
--- a/src/lib/stores/tokens/tokens.store.ts
+++ b/src/lib/stores/tokens/tokens.store.ts
@@ -22,6 +22,11 @@ export default (() => {
   let chainId: number | undefined;
   const tokenList = writable<TokenInfoWrapper[] | undefined>();
 
+  /**
+   * Connect the store to a chain, which is required before any other
+   * functionality can be used.
+   * @param toChainId The ID of the chain to serve tokens for.
+   */
   function connect(toChainId: number) {
     chainId = toChainId;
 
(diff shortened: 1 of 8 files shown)


efstajas:
Adding JSDocs

src/lib/stores/tokens/tokens.store.ts
```diff
@@ -0,0 +1,211 @@
...
+import * as storedTokens from './stored-custom-tokens';
+import assert from '$lib/utils/assert';
+import { Utils } from 'radicle-drips';
+
+interface DefaultTokenInfoWrapper {
+  info: TokenInfo;
+  source: 'default';
+}
+
+export interface CustomTokenInfoWrapper {
```
jtourkos, line 13: Since this is also used in the custom token store, I might consider extracting it out into a separate file.

src/lib/utils/assert.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Asserts that `condition` is truthy.
+ * @param condition The condition to assert.
+ * @param message Error message to throw on assertion error.
+ * @throw An error if `condition` is falsy.
+ */
+export default function assert(condition: unknown, message = 'Assertion Error'): asserts condition {
```
jtourkos, line 7: +1 for assertions! 🤘I might consider renaming in case we'll need to add more in the future (assertNotNull, assertNotNullOrEmpty, etc.), but we can always do it later.

jtourkos (approved):
Nice! 
Just a couple of minor suggestions (no need to make any changes).

src/lib/stores/tokens/tokens.store.ts
```diff
@@ -0,0 +1,211 @@
+import uniswapTokenList from '@uniswap/default-token-list';
+import type { TokenInfo } from '@uniswap/token-lists';
+import { get, writable } from 'svelte/store';
+import * as storedTokens from './stored-custom-tokens';
+import assert from '$lib/utils/assert';
+import { Utils } from 'radicle-drips';
+
+interface DefaultTokenInfoWrapper {
+  info: TokenInfo;
+  source: 'default';
```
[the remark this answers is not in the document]
efstajas: I generally prefer string literals as they functionally are pretty much identical to enums, but don't require importing the enum everywhere you want to use any of its values, which I've always found to be a bit boilerplate-y. Not a strong opinion on this though :)

src/lib/stores/tokens/tokens.store.ts
efstajas: 👍 If you don't mind, I'd do this when / if we add at least one or two more types — only reason I didn't already is that I thought having a types file and needing to export even the private types wasn't worth it for only 1 public definition.

PR #36, opened 2022-10-25
add Splits Section to Dashboard, closes #23

### LINKED ISSUE #23: Splits section
The Splits section displays the current account's splits configuration in a visual representation by default. The user may also switch over to a table list view of their splits. If a lot of splits are present, or some splits are too small to be represented in the visual bar view, it automatically switches to the table list view.

Designs: https://www.figma.com/file/vyI7f996JF8zwhnXwAwXdC/Drips-App-V2?node-id=169%3A5354

Image

evvvritt:
- styled like Figma
- no store / fetches inside component based on userId prop (so we can use the section on profiles)
- updated Dashboard to handle wallet-account changes

new to typescript + svelte so all feedback appreciated!

closes #23

### DIFF
--- a/src/routes/dashboard/+page.svelte
+++ b/src/routes/dashboard/+page.svelte
@@ -1,9 +1,23 @@
 <script lang="ts">
   import Balances from './sections/balances.section.svelte';
   import Streams from './sections/streams.section.svelte';
+  import Splits from './sections/splits.section.svelte';
 
   import { goto } from '$app/navigation';
   import wallet from '$lib/stores/wallet';
+  import { getAddressDriverClient } from '$lib/utils/get-drips-clients';
+
+  let userId: string;
+
+  async function getMyUserId() {
+    userId = (await (await getAddressDriverClient()).getUserId()).toString();
+  }
+
+  getMyUserId();
+
+  $: {
+    $wallet.address, getMyUserId();
+  }
 
   $: {
     if (!$wallet.connected) {
@@ -21,6 +35,7 @@
   <h1>Dashboard</h1>
   <Balances />
   <Streams />
+  <Splits {userId} />
 </div>
 
 <style>
--- a/src/routes/dashboard/sections/splits.section.svelte
+++ b/src/routes/dashboard/sections/splits.section.svelte
@@ -1,0 +1,94 @@
+<script lang="ts">
+  import type { SplitsEntry } from '$lib/stores/splits/types';
+  import type { UserId } from '$lib/stores/streams/types';
+  import MergeIcon from 'radicle-design-system/icons/Merge.svelte';
+  import PenIcon from 'radicle-design-system/icons/Pen.svelte';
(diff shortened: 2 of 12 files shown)


src/routes/dashboard/+page.svelte
```diff
@@ -20,9 +33,13 @@
...
-  <Balances />
-  <Streams />
-  <Splits />
+  {#await loading}
+    loading...
+  {:then}
+    <Balances />
+    <Streams />
+    <Splits {userId} />
+  {/await}
```
efstajas, line 42: Could the splits section accept an `undefined` `userId` and internally wait for it to be populated? You could pass `loaded = false` to `SectionSkeleton` in order to display a loading state only for the Splits section. That would avoid having another loading state above the sections, which already each have their own section-level loading state.

efstajas (changes requested):
Looks awesome! just one thing:

src/routes/dashboard/+page.svelte
evvvritt: Ok yep, [removed](https://github.com/radicle-dev/drips-app-v2/pull/36/commits/396bc452780a1c989116d712ceb3e994c070d3b9)

src/routes/dashboard/sections/splits.section.svelte
```diff
@@ -1,30 +1,56 @@
...
+  let splitsData: SplitsEntry[] | void;
+  let error = false;
+
+  onMount(() => userId && getSplits(userId));
+
+  $: {
+    getSplits(userId);
+  }
+
+  $: splitsTableData = formatSplitsTable(splitsData || []);
```
efstajas, line 24:
```suggestion
  $: splitsTableData = formatSplitsTable(splitsData ?? []);
```

🤓

src/routes/dashboard/sections/splits.section.svelte
```diff
@@ -40,15 +66,18 @@
...
-    <SectionSkeleton
-      emptyStateHeadline="No splits"
-      emptyStateEmoji="🫗"
-      emptyStateText="Anyone you split incoming funds with will appear here."
-      loaded={true}
-      empty={splitsData.length === 0}
-    >
-      <SplitsTable splits={splitsData} />
-    </SectionSkeleton>
+    {#key userId}
```
efstajas, line 69: what's the key for? if `userId` changes, it would change `splitsTableData`, which would cause the component to re-render anyway?

evvvritt: ah i was thinking to show the skeleton's loading state but i guess that's triggered in the getSplits function when setting to undefined before fetching

evvvritt: actually, going to keep key for now. skeleton container doesn't reset the height to the initial loading height

evvvritt: added a reset to initial height to skeleton when reloading

efstajas (approved):
🔥

PR #48, opened 2022-10-31
feat: validating input-address component. use in input-details stream…

evvvritt:
input component for address/ens validation (adapted from stream flow)


efstajas (approved):
Nice 👍 lot better to have this extracted as a component. Would be awesome if we could have a unit test for the validation logic… Maybe in another pr later :)

PR #52, opened 2022-11-01
fix: quick promises break stepper height

### LINKED ISSUE #51: If an await step resolves very quickly, it sometimes results in the next step not having the right height
Appears to be related to bind:this still pointing at the previous step that's being animated away.

efstajas:
Resolves #51 

We were checking the height of `stepElement`, which is not always reliably equal to the current step due to an apparent bug (?) in Svelte with `bind:this` on transitioned elements. Checking the height of `observedElement` instead appears to fix the problem.


evvvritt (approved):
works when i stage it in my splits flow :)

PR #54, opened 2022-11-01
fix: table visual alignment is bad

efstajas:
Adds the ability to set table column widths as percentages by setting `size` on a `ColumnDef`, and sets some percentages for the existing tables so that columns are properly visually aligned across the dashboard. 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ static/fonts/*
 !static/fonts/.gitkeep
 static/twemoji/*
 !static/twemoji/.gitkeep
+/coverage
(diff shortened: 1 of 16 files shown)


evvvritt (changes requested):
WIP
1. lets revise Balances and Streams column widths to below
1. and add Token column to end of streams table!

Screen Shot 2022-11-04 at 18 15 44

efstajas:
@evvvritt Updated :) 
Screenshot 2022-11-08 at 10 54 26

evvvritt (approved):

PR #102, opened 2022-11-16
feat: search

efstajas:
Adds search functionality for connected & disconnected browsing, plus E2E test.

Search can find streams, tokens, and profiles. It also allows jumping to Drips User IDs and unresolved ens names.

Since we don't yet have streams and token pages, these results currently just lead to a 404 page.


evvvritt (changes requested):
very nice!
but add`networkname` to forthcoming token route:
`/app/tokens/[networkname]/[token]`

evvvritt (approved):

PR #104, opened 2022-11-17
feat: page name for welcome page, homepage, improve social share values

efstajas:


src/routes/+page.svelte
```diff
@@ -13,6 +13,14 @@
   import DripsLogo from '$lib/components/header/drips-logo.svelte';
 </script>
 
+<svelte:head>
+  <title>Radicle Drips</title>
```
evvvritt, line 17: Drips

src/routes/app/+page.svelte
```diff
@@ -11,6 +11,11 @@
   }
 </script>
 
+<svelte:head>
+  <title>Radicle Drips - Welcome</title>
```
evvvritt, line 15: Welcome – Drips

evvvritt (changes requested):
@lftherios - do you have thoughts on meta/social Title and Description?

I was thinking the following:
**Drips**
**An Ethereum protocol for streaming and splitting funds, built by Radicle.**

or should Radicle be in the Title always?
**Radicle Drips**
**Drips by Radicle**

cc @brandonhaslegs

brandonhaslegs:
Why we welcoming someone to the website? seems very 2001 welcome to my homepage i'm a webmaster to me.

efstajas:
> Why we welcoming someone to the website? seems very 2001 like "welcome to my homepage i'm the webmaster".

what should we call the page instead?

lftherios:
> @lftherios - do you have thoughts on meta/social Title and Description?
> 
> I was thinking the following: **Drips** **An Ethereum protocol for streaming and splitting funds, built by Radicle.**

Yeah that's perfect!

brandonhaslegs:
> what should we call the page instead?

Just `Radicle Drips`, or just `Drips` on the first page. When visiting another page, you can just append the page name to that with a `|` or a `•` between.

efstajas:
I'd just argue that the page name is more important than `Drips` on nested pages, also imo just `Drips` is too short of a title for the landing page considering that it'll be indexed with that on Google. How about this:

```
- /: `Drips • Stream and Split Funds on Ethereum`
  - /app: `Drips`
    - /[userId]: `< user ID / ens > | Drips`
      - /[tokenAddress]: `< token name >  | Drips`
        - /[streamId]: `< stream name >  | Drips`
    - /dashboard: `Dashboard | Drips`
    - /settings: `Settings | Drips`
```

brandonhaslegs (approved):

evvvritt (changes requested):
change all titles to just "Drips"
otherwise it's redundant with description which appears with title on Google, social etc...
simpler + sexier!

efstajas:
@evvvritt Adjusted title for sharing on twitter etc. but kept same for Google, as discussed in the design chat 👍

evvvritt (approved):

PR #105, opened 2022-11-17
Token Page

evvvritt:
- adds Token route: `/app/:user/tokens/:token`*
- dynamic incoming/outgoing balances
- collect with tokenAddress preselected
- lists users' streams of that token

**to add later**
- ADD/WITHDRAW buttons active when flow is done
- remove Topup/Collect buttons on Dashboard when Add-Withdraw flow is done (+ add "Add Token" button)

tests needed? suggestions?


src/lib/stores/streams/streams.store.ts
```diff
@@ -108,6 +109,36 @@ export default (() => {
...
+
+        return {
+          totalEarned: acc.totalEarned + estimate.totalStreamed,
+          amountPerSecond: acc.amountPerSecond + estimate.currentAmountPerSecond,
+        };
+      },
+      { totalEarned: 0n, amountPerSecond: 0n },
+    );
+  }
+
```
efstajas, line 141: I think it'd be better to have this in the `balances` store to avoid tight coupling between `streams` and `balances`

src/lib/components/amount/amount.svelte
```diff
@@ -14,11 +14,14 @@
...
 
+  export let amountClasses = 'typo-text-mono-bold';
+  export let amountPerSecClasses = 'typo-text-small-mono text-foreground-level-4';
+
   $: amountTokenInfo = $tokens && amount ? tokens.getByAddress(amount.tokenAddress) : undefined;
   $: amountPerSecondTokenInfo =
     $tokens && amountPerSecond ? tokens.getByAddress(amountPerSecond.tokenAddress) : undefined;
 
-  function format(amount: Amount) {
+  export function format(amount: Amount) {
```
efstajas, line 24: don't think it needs to be exported?

```suggestion
function format(amount: Amount) {
```

src/routes/app/[userId]/tokens/[token]/+page.svelte
```diff
@@ -0,0 +1,137 @@
...
+  import collectFlowSteps from '../../../dashboard/sections/collect-flow/collect-flow-steps';
+  import modal from '$lib/stores/modal';
+  import wallet from '$lib/stores/wallet';
+  import guardConnected from '$lib/utils/guard-connected';
+
+  const urlParamToken = $page.params.token.toLowerCase();
+
+  $: token = $tokens?.find(
+    (token) =>
+      token.info.address.toLocaleLowerCase() === urlParamToken ||
```
Comment (efstajas) [inline src/routes/app/[userId]/tokens/[token]/+page.svelte:24]:
```suggestion
      token.info.address.toLowerCase() === urlParamToken ||
```

src/lib/components/table/table.svelte
```diff
@@ -38,8 +48,8 @@
     {/each}
   </thead>
   <tbody>
-    {#each $table.getRowModel().rows as row}
-      <tr>
+    {#each $table.getRowModel().rows as row, index}
+      <tr on:click={() => onRowClick(index)} class:cursor-pointer={isRowClickable}>
```
efstajas, line 52: it really would be great if we could get these to be `<a>` elements in order to preserve navigability and allow people to "open in new tab" etc. But I understand that that might require a larger refactoring of the table component. Would probably be good to log in an issue for this so we don't lose track.

efstajas (approved):
🤙 Looks great, few small suggestions but nothing blocking!

src/lib/components/table/table.svelte
evvvritt: yes agree, will make an issue

PR #123, opened 2022-11-24
feat: pause & unpause flows

### LINKED ISSUE #116: Pausing & deleting streams
Add the functionality to pause a stream (e.g. remove it from the current dripsConfig's receiver array, but leave it in Metadata) from the stream detail view. It should also be possible to "delete" a stream by removing it from metadata too.

efstajas:
Resolves #116

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "ethers": "^5.7.0",
         "fuzzysort": "^2.0.3",
         "radicle-design-system": "https://github.com/radicle-dev/radicle-design-system",
-        "radicle-drips": "github:radicle-dev/drips-js-sdk#v2",
+        "radicle-drips": "github:radicle-dev/drips-js-sdk",
         "web3modal": "^1.9.9",
         "zod": "^3.19.1"
       },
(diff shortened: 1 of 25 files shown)


efstajas:
Woops, discovered a lil bug, fixing

Screenshot 2022-11-25 at 11 00 06

evvvritt (approved):
works nicely :) but some observations:
1. button "Unpause" state should use play icon?
2. if i have one stream that's paused should the outgoing balance rate be 0.00?
3. should we show play/pause state in streams table
Screen Shot 2022-11-28 at 18 31 18

1 paused stream but outgoing rate not 0
Screen Shot 2022-11-28 at 18 29 19

efstajas:
@evvvritt 

> button "Unpause" state should use play icon?

👍 Fixed

> if i have one stream that's paused should the outgoing balance rate be 0.00?

👍 Fixed

> should we show play/pause state in streams table

Ya, I think so. Would it be ok for you if we make it a post-mvp feature tho? Since it would require a bunch of design

evvvritt:
> Ya, I think so. Would it be ok for you if we make it a post-mvp feature tho? Since it would require a bunch of design

ya meant post MVP @efstajas

PR #125, opened 2022-11-25
feat: global advisory store & app init loading state

### LINKED ISSUE #57: Some kind of loading UI while the wallet is being initialized
Right now, the app only mounts content after the wallet store is fully initialized. Sometimes, this process includes asking the user to switch to the right network through their connected wallet. If the user for some reason misses this request, they're just looking at a blank screen in the app, not knowing why. We should have some kind of UI while the app initializes, and also show "please switch network to x" while a network change request is pending.

efstajas:
Resolves #57

Adds a new "global-advisory" store, which can be used to show full-screen, blocking warnings, or a blue screen of death. Also replaces the blank screen while a wallet connection is restoring on first load with a spinner.

### DIFF
--- a/src/lib/components/global-advisory/global-advisory.svelte
+++ b/src/lib/components/global-advisory/global-advisory.svelte
@@ -1,0 +1,108 @@
+<script lang="ts">
+  import { browser } from '$app/environment';
+  import type {
+    FatalGlobalAdvisory,
+    NonFatalGlobalAdvisory,
+  } from '$lib/stores/global-advisory/global-advisory.store';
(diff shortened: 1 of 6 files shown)


evvvritt (approved):
Looks good. Would be good to add an animation to the header (if related). Blinks in rather than animates in like the end of init loading state.

efstajas:
> Looks good. Would be good to add an animation to the header (if related). Blinks in rather than animates in like the end of init loading state.

Ya, true, looks a bit rough. Gonna add a fix for that.

PR #126, opened 2022-11-25
feat: tooltip headers in balances section

### LINKED ISSUE #80: Add tooltips explaining streamable & earnings balances
It's pretty unclear what these are at the moment. Tooltips (like in the design) next to the two columns would help explain.

image

efstajas:
Adds explanatory tooltips for "Incoming" and "Outgoing" balance headers. Copy TBD

Screenshot 2022-11-25 at 16 01 26

Resolves #80

### DIFF
--- a/src/lib/components/table/table.svelte
+++ b/src/lib/components/table/table.svelte
@@ -8,7 +8,9 @@
   import { createSvelteTable, flexRender, type TableOptions } from '@tanstack/svelte-table';
   import ChevronDown from 'radicle-design-system/icons/ChevronDown.svelte';
   import ChevronUp from 'radicle-design-system/icons/ChevronUp.svelte';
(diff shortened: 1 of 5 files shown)


evvvritt (changes requested):
can you add to the token-page balance blocks too? 😍 
and maybe make a central file for the copy?

Screen Shot 2022-11-29 at 16 37 00

efstajas:
> can you add to the token-page balance blocks too? 😍 and maybe make a central file for the copy?

Added to token page 👍 

Re central file — almost feels like we should just generally switch to something like [svelte-i18n](https://github.com/kaisermann/svelte-i18n) so that we have all the copy in one place, rather than having only shared keys in some JSON file. If you don't mind I'd create an issue for that to do that post-MVP!

evvvritt (approved):

PR #127, opened 2022-11-25
feat: scheduled streams

### LINKED ISSUE #47: Improved stream end date input in Create Stream flow
In order to avoid confusion about the end date being translated to a duration relative to the stream creation timestamp (which we cannot accurately predict upfront), let's change the create stream flow to accept a specific end date only in combination with a specific stream start date. This ensures that the end date is always relative to the specified start date, avoiding the stream end date shifting relative to the transaction confirmation timestamp. In addition, a warning notifies the user that the stream may start after the configured start date, if said date is before the transaction's on-chain confirmation timestamp.

Image

efstajas:
Resolves #47 

Adds the ability to create scheduled streams with set start- and end dates.

Merge after #123

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "ethers": "^5.7.0",
         "fuzzysort": "^2.0.3",
         "radicle-design-system": "https://github.com/radicle-dev/radicle-design-system",
-        "radicle-drips": "github:radicle-dev/drips-js-sdk#v2",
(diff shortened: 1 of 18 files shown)


evvvritt (commented):
bug?
1. scheduled stream while out of funds
2. topped-up after stream ended

stream had "ended" label but animated drips flying, and "total streamed" accumulated for the duration (1m)

PR #140, opened 2022-11-29
feat: stepper visual improvements

efstajas:
- Restores pretty step height animation
- Hides step overflow while transitioning


evvvritt (commented):
Edit Splits doesn't animate height after loading splits?

efstajas:
> Edit Splits doesn't animate height after loading splits?

Ah, i see what you mean. The stepper only animates height changes between steps, while transitioning — i had a version in between that would animate all height changes, but it looked super glitchy when for example a text field validation error would pop up. For the edit splits modal I'd have two ideas: 

1. fetching splits with an `await` step as the first step and then passing it to the edit step as state (similar to `collect` flow).
2. decoupling the "monitor height and update with animation" logic that we already have in `section-skeleton` and `toggleable` into a generic component that just animates height changes, and re-using this within the edit splits step.

evvvritt:
> 1. fetching splits with an `await` step as the first step and then passing it to the edit step as state (similar to `collect` flow).

this is nice, let's add as an issue?

efstajas:
https://github.com/radicle-dev/drips-app-v2/issues/154

evvvritt (approved):

PR #149, opened 2022-12-01
fix: token amount input crashes when too many decimals

### LINKED ISSUE #122: Restrict amount of decimals for token amount inputs to the token decimals
Currently, ethers throws `Error: fractional component exceeds decimals` in all token amount inputs that we parse using `parseUnits` if the user enters more decimals than the token can handle. We should add the ability to set `step` on `TextInput` to the decision system so that we can restrict the amount of permitted decimals depending on the given token's precision.

efstajas:
Resolves #122

### DIFF
--- a/src/lib/flows/create-stream-flow/input-details.svelte
+++ b/src/lib/flows/create-stream-flow/input-details.svelte
@@ -6,7 +6,6 @@
   import TextInput from 'radicle-design-system/TextInput.svelte';
   import balances from '$lib/stores/balances';
   import tokens from '$lib/stores/tokens';
-  import { ethers } from 'ethers';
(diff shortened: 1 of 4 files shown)


evvvritt (approved):
IDEALLY it says something more descriptive to the issue. (eg "Exceeds maximum allowed decimals for this token.")

efstajas:
> IDEALLY it says something more descriptive to the issue. (eg "Exceeds maximum allowed decimals for this token.")

Yeah. I kept it vague because I suppose it's not necessarily guaranteed that the ethers function threw an error specifically for that reason. IMO it's enough of an edge case that keeping the error message vague is OK, while avoiding it being wrong. Wdyt?

PR #211, opened 2022-12-07
feat: tweaked modal transition

efstajas:
I think this feels a lot smoother.


evvvritt (approved):
tweaked it a little further:
bg fades in quickly first. 
was bothering me how the modal content cross-faded with the page content.

PR #220, opened 2022-12-09
feat: make all h1s the serif font

brandonhaslegs:
image
image
image
image

Signed-off-by: Brandon Oxendine <brandon.oxendine@gmail.com>


efstajas:
@brandonhaslegs should we add a secondary fallback to `serif`? Would look slightly better while the font is loading I guess

brandonhaslegs:
Yep, definitely.

efstajas (approved):

PR #222, opened 2022-12-12
feat: make clickable table rows tabbable

efstajas:
Allows tab-navigating through clickable table rows, and open the links in a new tab by holding `ctrl` or `command` while clicking. Also adds a prettier focus state to `identity-badge`.


evvvritt (changes requested):
1. `ctrl + click` identity badge open stream page in new tab
1. `ctrl + enter` on Identity Badge opens identity in current tab and opens stream page in new tab
2. `ctrl + enter` on stream row opens stream in current tab

efstajas:
@evvvritt should be all fixed 👍

evvvritt (approved):
I think it would be good to have a hover state for identity badge (especially in table row) so it's clear that takes to profile.

PR #228, opened 2022-12-12
feat: edit stream

### LINKED ISSUE #217: Edit Stream
Allow initially to change:

- the stream rate
- the stream name

All other fields should probably be unchangable for now, as changing them could be quite confusing.

efstajas:
Resolves #217


evvvritt (approved):
nice job!

PR #233, opened 2022-12-14
feat: Custom Tokens

### LINKED ISSUE #115: Custom tokens
Includes adding a new "Custom Tokens" page in settings, which should be a table displaying all `source: custom` tokens from the token store, and allow adding and removing custom tokens. Also includes adding entry-points to this page from "Unknown tokens" across the experience, should the user have topped-up or received some unknown token before adding it explicitly.

efstajas:
Screenshot 2022-12-15 at 13 27 15

Adds

-  a new "Custom tokens" setting, including a table that allows adding / removing custom tokens
- error messages for when you're trying to view a token or stream view with an unknown token, including ability to pre-fill the custom token address into the "add token" modal
- a little warning badge on custom tokens across the UI
- visual fixes for table rows not being the same size if they include an unknown token
- A new `sidestep` mechanism on `Stepper`, which allows temporarily moving to a different flow of steps as part of an original flow.
- The ability to add a new custom token straight from the "Add funds" flow.
- E2E test for adding a custom token from within the "Add funds" flow.
- A "Cancel" button to first steps of all flows.

Resolves #115


evvvritt (approved):
nicely done! I edited the warning texts in the modal and page a bit.

PR #234, opened 2022-12-15
fix: mobile search button not updated to new style

efstajas:
Screenshot 2022-12-15 at 13 33 03

Updates the mobile search button to match the new style. Also adds a 1px border around avatars.


evvvritt:
don't love the border on the 🔍, feel like it was fine without it (or perhaps a slight bg-color as we discussed as possible second button theme) @brandonhaslegs you have an opinion?

brandonhaslegs:
I agree it was fine the way it was!

PR #244, opened 2022-12-16
fix: #243, fix: #246

### LINKED ISSUE #243: "Edit stream" loses precision on amtPerSec
Currently, the "Edit Stream" modal simply takes the current amtPerSec on the stream, removes the extra digits of precision, and uses this as the / second value. This leads to the edit button unlocking immediately in most cases, and making it impossible to edit ONLY the name without overwriting the amtPerSec value with less precision.

We should try fitting the amtPerSec value with extra precision into the input by initially selecting an "Amount per" divider value higher than "/ second".

### LINKED ISSUE #246: Editing stream rate while stream is paused unpauses it
We should probably prevent editing a paused stream's stream rate.

efstajas:
Resolves #243
Resolves #246

This PR allows inputting extra-precise values into the amount fields (for both new streams and editing existing streams). The value is validated such that it always results in streaming at least 1 wei / cycle.

The amtPerSec input in the edit stream modal is also now locked if the stream is paused, so that editing doesn't unexpectedly unpause the stream.

image


efstajas:
Potential problem: When choosing a divider greater than / second, you consequently do NOT have the entire extra precision space available to you. Will need a slightly more complicated solution here.

PR #274, opened 2023-01-12
feat: make all illustration assets the same size

brandonhaslegs:
Might need to adjust the size of the SVGs now, they look a bit small.

Signed-off-by: Brandon Oxendine <brandon.oxendine@gmail.com>


efstajas:
Thanks! Included in #267

PR #294, opened 2023-01-30
feat: 3d button

efstajas:


brandonhaslegs:
I still think that the buttons should have no shadow on press-down state. That's how a 3D button works. @evvvritt wyt?

evvvritt:
agree with @brandonhaslegs. also, not sure about :focus being the raised button state 🤔 but maybe makes sense? feels like with the effect on hover, when you close a modal and the button is raised the ui like fell asleep

PR #323, opened 2023-02-27
feat: Safe app compatibility

### LINKED ISSUE #112: Improved experience with Gnosis Safes
Currently, the app should be fully usable with a Gnosis Safe via the WalletConnect Safe App, but unfortunately the experience won't be great, as the app will expect a transaction to resolve short-term. We should consider detecting that the app is connected to a Gnosis Safe, and merely propose transactions to the safe, rather than expecting them to resolve immediately.

### LINKED ISSUE #273: Improved Gnosis Safe support
The app can already be used through the WalletConnect safe app, but the experience isn't ideal — the app doesn't realize that it's talking to safe, and will wait for transactions to be confirmed synchronously. This obviously doesn't work for when you're using a safe.

We should detect if the app is connected to a safe, and slightly alter all transaction flows to treat the transaction as asynchronous operations.

### LINKED ISSUE #326: Safe App: Add notices for time-sensitive transactions
When connected to a safe, we should add some extra warnings to several interactions that are time-sensitive, such as

- scheduling a stream
- withdrawing funds from outgoing balance
- pausing & unpausing a stream

Additionally, we should warn the user somewhere (docs?) about transactions that depend on history failing if a previous transaction was scheduled before and executed after a second transaction is proposed to the safe.

efstajas:
Resolves #273 
Resolves #112 
Resolves #326


efstajas:
I noticed that `is-running-in-safe` doesn't appear to be working reliably on the deployed version. Investigating…

efstajas:
fixed

efstajas:
Found a problem with the "add custom token" flow showing the "continue in your safe" message at the end. doing some refactoring to fix

efstajas:
E2E pass ✔️

evvvritt (approved):
Nice work!!

Added a commit for what I feel is clearer language for the disclaimer (felt a bit german -> english, no offense ;)

Successfully Tested: Add Funds, Edit Splits, Create Stream, Collect, Withdraw
in the gnosis safe app interface

One nice to have would be a label for this action (during Approve/AddFunds)
Screenshot 2023-03-06 at 15 54 04

efstajas:
@evvvritt thx but the new disclaimer actually isn't accurate anymore - if there already any pending transactions right now and they DO get executed, the new transaction will fail because it was planned with a now-outdated drips- or splits config. If the currently-pending transaction is cancelled, the new transaction would execute fine.

PR #334, opened 2023-03-06
feat: #329

### LINKED ISSUE #329: Prevent navigating away from page while flow is open
Currently, it's pretty easy to accidentally press the back button or close the window while entering data in the flow, which could result in an annoying loss-of-input. We should use the browser's native "are you sure you want to leave? you might lose data" dialog in these cases.

efstajas:
Prevents navigation away from page while a modal is open.

Resolves #329

### DIFF
--- a/src/lib/stores/modal/modal.store.ts
+++ b/src/lib/stores/modal/modal.store.ts
@@ -47,6 +47,8 @@ export const hide = (): void => {
 
   stored.overlay && stored.overlay.onHide();
   overlayStore.set(null);
+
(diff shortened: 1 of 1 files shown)


evvvritt:
If I press browser back button while a modal is open I don't get the confirmation guard? (part of the issue)

efstajas:
You do, but only if the navigation would actually result in the page reloading or disappearing fully. If it just navigates between pages within the app, it does allow the navigation, but the modal also stays open

evvvritt (approved):

Issue #94, opened 2022-11-14
Allow social share links like drips.network/[ens]


efstajas:
Our profile view is currently at /app/[ens]. Would be cool to make it just \[ens] to get the link to be as short as possible.

PR #347, opened 2023-03-09
feat: #94

efstajas:
Redirects /[user-id] to /app/[user-id]


evvvritt:
What's blocking us from getting rid of /app/ altogether again?

efstajas:
> What's blocking us from getting rid of /app/ altogether again?

With the "app" scoped under /app we need just a single file (/app/+layout.svelte) to load all stores (wallet, tick etc.) and apply the app's layout to all routes nested below. It'd be a lot more complicated to do that if we have non-app routes (like the upcoming legal pages and probably some new marketing pages at some point) and app routes mixed under /.

PR #349, opened 2023-03-09
feat: save & restore input values in all flows

### LINKED ISSUE #192: Keep the data in the schedule form inputs when retrying
When I tried to set a scheduled stream and failed (possibly due to the known gas issue), after I clicked the "try again", I'd expect the start/end datetimes to be there as before.

efstajas:
Automatically saves & restores form inputs within stepper flows after coming back to a step (e.g. after an error or declining a transaction and clicking "try again")

Resolves #192


evvvritt (approved):
love it!

PR #356, opened 2023-03-10
feat: state badges on dashboard

### LINKED ISSUE #179: Flag paused streams
The list of drips isn't clear about the status of paused streams, it just displays a zero dripping rate, which is technically true, but inconsistent with the mental model of pausing things. It'd be cool to keep displaying the target rate but explicitly mark it as paused.

image

efstajas:
Resolves #179 

Screenshot 2023-03-10 at 14 36 01

### DIFF
--- a/src/lib/components/stream-state-badge/stream-state-badge.svelte
+++ b/src/lib/components/stream-state-badge/stream-state-badge.svelte
@@ -3,7 +3,49 @@
 </script>
 
 <script lang="ts">
-  export let state: StreamState;
+  import balances from '$lib/stores/balances/balances.store';
+  import streams from '$lib/stores/streams';
+  import type { StreamId } from '$lib/stores/streams/types';
+  import getStreamHistory from '$lib/utils/stream-history';
+  import unreachable from '$lib/utils/unreachable';
+
+  export let streamId: StreamId;
+  export let paused: boolean;
+  export let durationSeconds: number | undefined = undefined;
+  export let startDate: Date | undefined = undefined;
+  export let senderId: string;
+  export let tokenAddress: string;
+
+  export let hideActive = false;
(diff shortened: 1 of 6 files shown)


evvvritt (commented):
(added a commit for style consistency)

I added funds on Token view. On confirm, the outgoing balance started moving again but the states of the streams list did not. 

Is it complicated to have it update on topup confirmation without needing to refresh the page?

Screenshot 2023-03-10 at 15 01 38

efstajas:
@evvvritt fixed

efstajas:
@evvvritt I deliberately made the badge the drip shape everywhere in the previous commit, also on the stream page. I thought it was weird that this badge was the only thing fully rounded

evvvritt:
It feels kinda overkill to me and I think it's fine we leave the drip shape to UI frames and buttons. Maybe @brandonhaslegs has an opinion?

brandonhaslegs:
Maybe the `Ended` badge is too much? Paused seems good though. Would have to really see what a real life scenario looks like. Maybe there's another way to indicate it's ended. Only reason I mention this is because Ended seems like it might be all over the place on a lot of streams.

efstajas:
@brandonhaslegs I think Ended is good to have as a badge - if you scheduled a stream to end at some particular point, you would really want to just delete it once it's "ended". It also only appears if you specifically set an end date for the stream.

Anyway, other question is whether you think that the state badge should be fully rounded, or rounded with the drip-shape.

efstajas:
fully round

Screenshot 2023-03-10 at 16 04 39

Drip shape

Screenshot 2023-03-10 at 16 05 19

brandonhaslegs:
I feel like go big or go home. Why would we stop doing the drip rounding now?

evvvritt (approved):

PR #360, opened 2023-03-13
feat: show "back to drips v1" card on profiles

efstajas:
Adds a dismissable "looking for drips v1?" card at the bottom of profiles that allows quickly opening the current profile on Drips V1.

Screenshot 2023-03-13 at 10 35 11

### DIFF
--- a/src/lib/components/banner/banner.svelte
+++ b/src/lib/components/banner/banner.svelte
@@ -25,7 +25,7 @@
     </div>
   </div>
   <div class="right">
-    <a href={button.href}>
+    <a href={button.href} rel="noreferrer" target="_blank">
       <Button variant="primary">
         {button.label}
       </Button>
--- a/src/lib/components/banner/banner.svelte
+++ b/src/lib/components/banner/banner.svelte
@@ -26,12 +26,12 @@
   </div>
   <div class="right">
     <a href={button.href}>
-      <Button>
+      <Button variant="primary">
         {button.label}
       </Button>
     </a>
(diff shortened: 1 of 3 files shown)


evvvritt:
Hmm, I think I prefer a yellow banner like on homepage? @brandonhaslegs wdyt
1 card feels odd, I guess maybe opportunity for other "dismissables" too but banner feels more conventional.

brandonhaslegs:
What was the reason for adding it?

efstajas:
@brandonhaslegs after the soft launch, beta.drips.network/address which previously led to the drips V1 profile will now lead to the Drips V2 profile. Given there may still be some profile links like that around out there that people are expecting to lead to V1, we thought it'd be good to add an easy way to switch straight to the same profile on V1.

@evvvritt banner makes sense to me, I'll update the PR later 👍

efstajas:
@evvvritt 

Screenshot 2023-03-14 at 12 03 15

efstajas:
just made it primary - anyway thought it looked too much like a bad thing in that warning color :)

Screenshot 2023-03-14 at 12 14 31

brandonhaslegs:
1. This is VERY prominent. Can it go under the meta data above the balances?

2. Does this show up on every profile or only ones that were previously on Drips1?

3. Maybe instead of "Drips V1" we could refer to it a bit more colloquially. Like, "The old Drips".

evvvritt:
Ya... I'd vote the very bottom even.

evvvritt (approved):
Tweaked the description a little but good to go :)

PR #372, opened 2023-03-14
fix: resolve #363 #364

### LINKED ISSUE #363: address-input assigns address value too early, results in a lot of invalid address errors being thrown
Screenshot 2023-03-13 at 15 03 28

See errors in console while typing into address field, also see that it prematurely assigns the address, resulting in the right side of the stream illustration displaying it too early.

### LINKED ISSUE #364: pre-populated address-input results in create stream form being locked even though it's valid
Screenshot 2023-03-13 at 15 04 51

Repro:
- Open create stream modal
- Populate the form with valid values
- Submit
- Decline transaction and click "Go back" on error modal
- Observe form is filled out completely with valid values, but "Create stream" button is locked

evvvritt:
resolves #363
resolves #364 

address input remains aggressive (validates on every input change) but I think this is best, otherwise the form can be locked without any user feedback.

### DIFF
--- a/src/lib/components/input-address/input-address.svelte
+++ b/src/lib/components/input-address/input-address.svelte
@@ -13,12 +13,7 @@
     msg: string;
   };
 
-  export let exclude: ExclusionGroup[] = [
-    {
-      addresses: [],
-      msg: 'You cannot use this address.',
-    },
-  ];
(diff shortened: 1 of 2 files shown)


efstajas (approved):
🎉

PR #399, opened 2023-04-07
feat: new base layout & navigation

efstajas:
- New /app/crowdfunding stub route.
- Move "dashboard" to /app/streams
- New global sidenav for desktop
- New global bottom nav for mobile
- New account menu bottom sheet for mobile
- New header layout with search and settings buttons


src/routes/app/streams/+page.svelte
```diff
@@ -124,12 +124,11 @@
 </script>
 
 <svelte:head>
-  <title>Dashboard | Drips</title>
+  <title>Streams | Drips</title>
```
jtourkos, line 127: Should this be dynamic depending on the current page (streams vs. crowdfunding)?

jtourkos (commented):
Added a minor comment in the code.

And an edge case:
- In desktop mode select the streams tab
- Resize to mobile
- The correct tab is still selected
- Select back and forth a couple of times streams and crowdfunding and end up to streams.
- Resize again until you see the side nav
- The crowdfunding button is selected where the streams should be.

No one is going to resize the windows like this, but adding it since I noticed.

Other than that, great job, all looks good to me!👌🚀

efstajas:
@jtourkos good catch. I'll just make it reposition the selector on window resize.

efstajas:
I guess the dashboard link in the account menu makes no sense anymore, so I should remove that. That leaves only "Profile" in the account menu. 😬 I guess fine for now but we should think about redesigning it before release.

efstajas:
@evvvritt yeah totally, just noticed the same (see prev comment). I'll just remove it for now 👍

efstajas:
Ah, I forgot about adding tooltips on the collapsed sidenav. Will add to this PR as well.

jtourkos (approved):

brandonhaslegs (approved):
Please don't merge this on a Friday, if there's a bug our users will hate us.

PR #415, opened 2023-04-19
Metadata APIs

jtourkos:
Here's the first implementation of the new Metadata APIs. It's still a WIP, but I wanted to validate the approach.

Implementation was straightforward, but mocking with Vitest was more challenging than any other testing framework I previously worked with. But, all good now, up to speed.

The main idea is:
- We have 3 classes for managing metadata: `NftDriverMetadataManager`, `AddressDriverMetadataManager` and `GitDriverMetadataManager`
- All 3 inherit from a `MetadataManagerBase` that implements:
    - `fetchMetadataHashByUserId`
    - `fetchAccountMetadata`
    - `pinAccountMetadata`
    - `updateAccountMetadata`
    and handles the schemas
- Each metadata manager is enhanced with specific methods. For example, `` also exposes `fetchAccount` and `generateAccountMetadata`


.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
...
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run-script", "test"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Vite Tests by Pattern",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run-script", "test", "MetadataManagerBase"]
+    },
```
efstajas, line 17: There's an excellent Vitest extension available for VSCode FYI.

PR #445, opened 2023-05-03
feat: fiat estimation utilities

efstajas:
Adds new fiat estimate utilities based on Binance's market data WS streams.


src/lib/utils/deduplicate-readable.ts
```diff
@@ -0,0 +1,17 @@
...
+export default function <T>(store: Readable<T>): Readable<T> {
+  let previous: T;
+
+  return derived(store, ($value, set) => {
+    if (Array.isArray($value) && Array.isArray(previous)) {
+      if ($value.toString() === previous.toString()) return;
+    }
+
+    if ($value !== previous) {
+      previous = $value;
```
jtourkos, line 13: This would set `previous` and `$value` to hold a reference to the same object in memory. If `$value` changes elsewhere, `previous` would also change. Is this a valid scenario?

src/lib/utils/fiat-estimates/fiat-estimates.ts
```diff
@@ -0,0 +1,202 @@
...
+  symbol = TOKEN_SUBSTITUTIONS[symbol] || symbol;
+  _validateSymbol(symbol);
+
+  const price = get(prices)[symbol];
+
+  if (!price) return undefined;
+  if (typeof price === 'string') return price;
+
+  return (
+    (parseFloat(price.toString()) * parseFloat(amount.amount.toString())) /
```
jtourkos, line 176: Why `parseFloat`? Aren't `price` and `amount.amount` already `number`s?

src/lib/utils/fiat-estimates/fiat-estimates.ts
```diff
@@ -0,0 +1,202 @@
...
+    const price = parseFloat(data.c);
+
+    prices.update((prices) => ({
+      ...prices,
+      [symbol]: price,
+    }));
+  }
+}
+
+function _validateSymbol(symbol: string): asserts symbol is SupportedSymbol {
```
jtourkos, line 48: Would we need to `try-catch` the error it might throw on the caller's side?

jtourkos (approved):
👌👌

src/lib/utils/deduplicate-readable.ts
efstajas: I don't think so, since `$value` only exists within the scope of this callback, and nothing in this callback alters it.

jtourkos: Oh, right. I somehow confused it that it is part of the "store" and could be changed elsewhere.

src/lib/utils/fiat-estimates/fiat-estimates.ts
```diff
@@ -0,0 +1,202 @@
+import { derived, writable, get } from 'svelte/store';
+import assert from '$lib/utils/assert';
+import deduplicateReadable from '../deduplicate-readable';
+import tokensStore from '$lib/stores/tokens/tokens.store';
+import createSocket from './sockets';
+import { BinanceCommand, BinanceMessage } from './binance-ws-api/types';
+import { SUPPORTED_SYMBOLS, type SupportedSymbol } from './binance-ws-api/supported-symbols';
+import type { z } from 'zod';
+
+interface Amount {
```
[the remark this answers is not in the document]
efstajas: Oh ya. Let's do that in a separate PR then and for all `Amount` types, because this isn't the only place (and i'd rather have them all be the same so you can pass them around).

PR #466, opened 2023-06-16
Add fuder-onboarding transactions

jtourkos:


src/lib/utils/get-drips-clients.ts
```diff
@@ -120,7 +120,7 @@ export function getCallerClient() {
 export const networkConfigs: { [chainId: number]: NetworkConfig } = isTest()
   ? {
       5: {
```
efstajas, line 122: should be the chain ID of sepolia?

efstajas (changes requested):
Just one little thing ⬇️

src/lib/utils/get-drips-clients.ts
jtourkos: Good catch, changed.

PR #475, opened 2023-06-28
feat: support nft receivers for streams store, display drip list streams 

efstajas:


jtourkos (approved):
I assume this will only break if someone starts to experimenting directly with the protocol. But for now we're fine.

PR #575, opened 2023-07-13
Projects without metadata as unclaimed

jtourkos:
resolves https://github.com/radicle-dev/drips-app-v2/issues/536

### DIFF
--- a/src/lib/utils/project/GitProjectService.ts
+++ b/src/lib/utils/project/GitProjectService.ts
@@ -628,7 +628,7 @@ export default class GitProjectService {
           accountId,
           driver: 'repo',
         },
-        verificationStatus: VerificationStatus.IN_PROGRESS, // Project is claimed here and has no metadata. We assume it's in progress.
+        verificationStatus: VerificationStatus.FINISHED,
         source: GitProjectService.populateSource(Number(onChainProject.forge), repoName, username),
       };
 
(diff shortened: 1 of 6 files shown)


src/lib/utils/project/GitProjectService.ts
```diff
@@ -622,29 +619,24 @@ export default class GitProjectService {
...
+    // ...and hasn't metadata. Return as unclaimed project.
+    if (!projectMetadata) {
+      const unclaimedProject: UnclaimedGitProject = {
+        claimed: false,
+        owner: undefined,
+        repoDriverAccount: {
+          accountId,
+          driver: 'repo',
+        },
+        verificationStatus: VerificationStatus.IN_PROGRESS, // Project is claimed here and has no metadata. We assume it's in progress.
```
efstajas, line 631: If I understand correctly, this is the case where we have an owner on-chain, but no metadata, right?

in that case, I think this should either be a new `VerificationStatus` like `FINISHED`, or `NOT_STARTED` if we want to just pretend that the claim process hasn't even started yet.

efstajas (changes requested):

src/lib/utils/project/GitProjectService.ts
jtourkos: Changed to `FINISHED` 👍

efstajas (approved):
LGTM 👍

Issue #629, opened 2023-08-14
Design QA


brandonhaslegs:
# Edit Drip List Modal
- [x] add Trash icon next to the percentage which removes the recipient
- [x] remove the checkbox functionality (this was always confusing)
- [x] Add eth address in small grey text under the ENS name
- [x] Change wording on buttons from "Distribute evenly" etc… to "Split evenly" etc…
- [x] Change primary button on modal to "Confirm changes in *your* wallet"

# Streams page
- [x] Remove the X in the top right of the Getting Started tips, add a ghost "Dismiss" button nexst to the main CTA
- [x] In the "Learn how to collect funds you’ve earned from incoming streams or splits." edu card, make the apostrophe a curly one not a straight one.
- [x] On the balance line item, the chevron button to the right should be foreground, not grey-4. ghost buttons should be foreground instead of grey-4 generally
- [x] In the Streams section, do we need the "↑ Outgoing" heading? What are the other states of this filter?

# Global
- [x] Disable the font style setting that makes the "x" character be positioned in the middle of the Cap height when it's between two numbers. i.e. 0x4 shows up as 0×4, when it should show up as 0x4.
- [x] Make "Cancel" button a ghost button on modals instead of a outline button
- [ ] #656
- [x] the chevron on the dropdown component should be foreground, not level 4

# Add funds modal
- [x] the token ticker symbol should be grey-4 instead of foreground
- [x] placeholder text should be "Search tokens" instead of just "Search"

# Add USDC funds modal
- [x] "Wallet Balance" balance shouldn't be capitalized
- [x] instead of "Go back" change text to "Back"

# Single Stream page
- [x] move the "state" tag next to the stream name (just like it is on the drip list stream page)
- [x] #655
- [x] change the h4 headline colors of "Remaining balance for token" and "balance runs out of funds" and "created at" to `foreground` so that they match the other h4 headlines on the page
- [x] instead of "Created at" it should say "Created on"
- [x] what's the difference between "Created on" and "started"?

# Delete stream modal
- [x] the text field with the stream name in it has incorrect rounded corners. Should be fully round

# Create a stream modal
- [x] The text "Specify start- and end dates" shouldn't have a hyphen.

# Profile
- [x] Change the meta data icons color to foreground instead of level-4

# Project page
- [x] bold the eth address before "is raising funds for" so that it matches the "created a Drip List" heading on the Drip List page.

# Account Dropdown
- [x] Make the disconnect button a ghost button. takes too much focus being a normal button
[...]

PR #631, opened 2023-08-16
Ev/drip list trash items

evvvritt:
from #629
> [ ] add Trash icon next to the percentage which removes the recipient
> [ ] remove the checkbox functionality (this was always confusing)
> [ ] Change wording on buttons from "Distribute evenly" etc… to "Split evenly" etc…

- list-editor no longer uses list-select.svelte
    - but currently does not clean up list-select.svelte (like removing PercentageEditor?)
    - also removes "selected" logic (just edits items/percents directly)
- maintains tabbing
- "distribute remaining" affects any items with 0% (like 0xSplits :)

@efstajas - maybe take a look and see if you think it's ready to merge? did some manual testing in funder + edit drip list flows


efstajas:
@evvvritt Could you check the `npm run check` output?

efstajas (approved):
Looks awesome, so much simpler. just one minor comment, and would be great if you could make sure the percentage & delete button are always visible on mobile

PR #638, opened 2023-08-28
feat: funder list stored locally

evvvritt:
- minimal implementation (using z.any() / no type checking)
- clears stored if local is corrupted or doesn't match state (cheap check)

should we remove the "changes may not be saved" page reload popup guard?
(if so, where is that?)


evvvritt:
ah, the e2e fails because we can't save the stream value as a bigint (fails when storedWriteable tries to stringify it)

i guess an idea is to just store the dripList as a storedWriteable copy of the steppers writeable state? and load it if it's present?
(refresh loads at the drip list editor anyways)

or is it important to save the stream inputs?

evvvritt:
@efstajas wdyt^?

PR #639, opened 2023-08-29
Ev/iss629 streams

evvvritt:
streams section of #629 
@efstajas, I can't assign brandon to review PRs?


src/e2e-tests/top-up-create-stream.e2e.test.ts
```diff
@@ -168,7 +168,6 @@ describe('app', async () => {
       it('displays the original users outgoing stream on their profile', async () => {
         await page.goto('http://127.0.0.1:3000/app/0x433220a86126eFe2b8C98a723E73eBAd2D0CbaDc');
 
-        await expect(page.locator('text=↑ Outgoing')).toHaveCount(1);
```
efstajas, line 171: Removing this kinda removes the entire check that this new stream is being interpreted by the streams store as outgoing, rather than incoming. 😬 Would be good to keep that

efstajas: maybe just check that _neither_ "incoming" or "outgoing" appear at this point. that should do it

efstajas (changes requested):
just one tiny lil thing

src/e2e-tests/top-up-create-stream.e2e.test.ts
evvvritt: In 3f2ed36, I check for "From" and "Total streamed" since outgoing/incoming = 0 would be true in any situation without both stream types present

evvvritt: g2g?

efstajas (approved):
LGTM 👍

PR #641, opened 2023-08-29
Ev/iss629

evvvritt:
selected items from #629 

## Global
- [ ]  Disable the font style setting that makes the "x" character be positioned in the middle of the Cap height when it's between two numbers. i.e. 0x4 shows up as 0×4, when it should show up as 0x4.
- [ ] Make "Cancel" button a ghost button on modals instead of a outline button
- [ ] the chevron on the dropdown component should be foreground, not level 4

## Add funds modal
- [ ] the token ticker symbol should be grey-4 instead of foreground
- [ ] placeholder text should be "Search tokens" instead of just "Search"


brandonhaslegs:
Can you also add a "Dismiss" button here and remove the X? Just like the other informational widgets across the app?

image

evvvritt:
yep, updated:
Screenshot 2023-09-07 at 15 31 03

brandonhaslegs (approved):

PR #661, opened 2023-09-07
style: fiat-est + token amount uses en-us locale (closes #386)

### LINKED ISSUE #386: Comma instead of period
image

This is only on the add funds modal, when you have "Max" selected

This should be a period.

And arguably if it's a round number it doesn't need any decimal at all.

evvvritt:
closes #386


efstajas (approved):
LGTM 👍

PR #670, opened 2023-09-14
truncate (...) long project name (closes #540)

### LINKED ISSUE #540: The repo name field doesn't do ellipsis
image

evvvritt:
closes #540
- for some reason display:grid breaks .truncate so using flex for project columns
- reflow project-header for mobile
- github link max-width = 320px
screenshots below (note, Edit button properly only applies for owner*)
Screenshot 2023-09-14 at 13 49 16
Screenshot 2023-09-14 at 13 48 36
Screenshot 2023-09-14 at 13 48 24
Screenshot 2023-09-14 at 13 49 28


evvvritt:
@brandonhaslegs wdyt of the design? 
- project name under avatar on mobile
- edit/customize button in upper right corner

evvvritt:
ah, need to fix one more truncate instance (in claim project modal slot)

brandonhaslegs:
I (annoyingly, sorry) don't think we should truncate the project name! it should flow to the next line. Otherwise you can't fully read it anywhere.

evvvritt:
Ok, I removed the truncate on the big project title in the header, but keeping it the badge. Good @brandonhaslegs?

Screenshot 2023-09-14 at 14 42 46
Screenshot 2023-09-14 at 14 42 08

brandonhaslegs (approved):

src/lib/components/stepper/stepper.svelte
```diff
@@ -198,6 +198,8 @@
       disableTransitions = false;
     } else {
       modal.hide();
+      // dispatch in case Stepper is being used in a standalone page (like /funder-onboarding)
+      dispatch('conclude');
```
efstajas, line 202: is this change related to the PR? idgi 👀

src/routes/app/(flows)/funder-onboarding/+page.svelte
```diff
@@ -31,6 +31,7 @@
 <Stepper
   bind:currentStepIndex
   on:stepChange={() => window.scrollTo({ top: 0 })}
+  on:conclude={() => window.history.back()}
```
efstajas, line 34: what if you got linked to the flow from somewhere else, e.g. google or a discord link?

src/routes/app/(flows)/funder-onboarding/steps/build-list/build-list.svelte
```diff
@@ -28,6 +29,9 @@
     showListFirst={true}
     {projectUrlToAdd}
   />
+  <svelte:fragment slot="left-actions">
+    <Button icon={ArrowLeft} on:click={() => dispatch('conclude')}>Back</Button>
+  </svelte:fragment>
```
efstajas, line 34: Design changes LGTM, but not so sure about this "back" button implementation using `conclude` and `window.history.back`;

1. user could be linked to this flow from anywhere, including external sites, in which case it's weird to have an on-page back button. we'd have no idea where it leads to
2. the flow isn't `conclude`d, so i don't think dispatching the `conclude` event is the right one
3. Should probably be a separate PR to this one, since it doesn't have anything to do with truncating the project names.
4. We should probably just trigger the flow in a modal. For that, we'd just need to move the two flows with dedicated pages to the `flows` dir, then it should be pretty trivial to put them into a modal like many other flows already are.

efstajas (changes requested):

src/routes/app/(flows)/funder-onboarding/steps/build-list/build-list.svelte
evvvritt: This code is already live on main, I must have accidentally branched this one off the back button issue branch.

I did window.history.back() because doing a goto('/') has a really gnarly page transition that I'm not sure why. While back() properly scrolls it to the place you were on the landing page. 

I was considering this a temporary solution. Sensible to make a new issue for it. I think the trouble I had was determining whether you came from landing page or entered the site at that flow. Using afterRoute() to detect had identical results.

evvvritt:
@efstajas, reverted to `display:grid` for project layout so the sidebar is back under the header on mobile (but with some improvements on laptop).

With the `widow.history.back()` being discussed in #684, lmk if everything else looks ok?

src/routes/app/(flows)/funder-onboarding/steps/build-list/build-list.svelte
evvvritt: (resolving since handled by #684)

efstajas:
@evvvritt could you check why the tests are failing?

efstajas (approved):
LGTM, lots of bigger layout changes but wasn't able to spot any problems!

PR #671, opened 2023-09-16
refactor: use coinmarketcap to generate static fiat estimates

efstajas:
It turns out that the Binance WS endpoint isn't accessible globally, so this PR changes our provider for fiat estimates to the Coinmarketcap API.

**Important**: Adds a new env variable `COINMARKETCAP_API_KEY`. To get one for local development, sign up for a free developer account with Coinmarketcap. To test this PR with mainnet, check out the branch, set .env value `PUBLIC_NETWORK` to `1`, and populate `COINMARKETCAP_API_KEY` with your own key.

This is largely re-using the existing API surface of the `fiat-estimates` util, in order to keep the door open to bring back real-time (or periodic) price updates in the future.

All Coinmarketcap API calls are proxied through our SvelteKit backend. On call to `fiatEstimates.start`, the app fetches a map of ERC-20 token addresses to Coinmarketcap IDs, which is a lot safer than relying on symbols, the uniqueness of which is not guaranteed. On call to `track`, it now fetches the current USD value of the given tokens once. There's currently no logic for updating currently-tracked tokens over time, but we can bring this back in the future while keeping the current API surface.


src/lib/utils/fiat-estimates/fiat-estimates.ts
```diff
@@ -2,180 +2,113 @@ import { derived, writable, get } from 'svelte/store';
...
-    method: 'UNSUBSCRIBE',
-    params: symbols.map((symbol) => `${symbol.toLowerCase()}usdt@ticker`),
-    id: 1,
+  // Update the prices store with the new prices
+  prices.update(($prices) => {
+    return {
+      ...$prices,
+      ...Object.fromEntries(
+        Object.entries(parsedRes).map(([id, price]) => [
+          utils.getAddress(addresses[ids.indexOf(parseInt(id))].toLowerCase()),
```
efstajas, line 107: Just noticed this has a chance to pick the wrong address if there are any unsupported addresses... Will fix!

efstajas: Fixed this 👍

jtourkos (commented):
Code looks good 👍

Just a thing I noticed while testing locally:

If you manually refresh the page (I was testing http://localhost:5173/app/projects/github/libgit2/libgit2), you sometimes get the estimation, but some other times you get:

```
Uncaught (in promise) Error: Token info not found for tracked token
    at assert (assert.ts:9:11)
    at Object.convert (fiat-estimates.ts:139:3)
    at token-amounts-table.svelte:37:49
    at Array.forEach (<anonymous>)
    at $$self.$$.update (token-amounts-table.svelte:36:13)
    at update (index.mjs:1343:12)
    at flush (index.mjs:1307:17)
 ```

**Comment by efstajas** [comment]
@jtourkos great catch, I think I know why it's happening. Will push a fix shortly

File: src/lib/utils/fiat-estimates/fiat-estimates.ts [hunk 2]
```diff
@@ -2,180 +2,120 @@ import { derived, writable, get } from 'svelte/store';
...
-import createSocket from './sockets';
-import { BinanceCommand, BinanceMessage } from './binance-ws-api/types';
-import { SUPPORTED_SYMBOLS, type SupportedSymbol } from './binance-ws-api/supported-symbols';
-import type { z } from 'zod';
+import { z } from 'zod';
 import { formatUnits } from 'ethers/lib/utils';
+import { utils } from 'ethers';
+
+type TokenAddress = string;
+type DataProviderTokenId = number;
```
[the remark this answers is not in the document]
efstajas: nice, never heard of this before!

jtourkos (approved):

PR #666, opened 2023-09-12
add 'back' to funder flow (closes #659)

evvvritt:
having Brandon review,
but @efstajas, holler if there's a certain reason the landing page CTAs were opening in a new tab (I've since removed)
closes #659

PR #684, opened 2023-09-25
Revert "fix: add back button to funder flow, first step (#666)"

efstajas:
Imo best to revert this for now:

- user could be linked to this flow from anywhere, including external sites, in which case it's weird to have an on-page back button. we'd have no idea where it leads to
- if you enter the URL directly or get linked from an external app (e.g. Discord), there's a "back" button, but it just doesn't do anything
- the flow isn't being "concluded", so i don't think dispatching the `conclude` event is appropriate

We should rather work on putting this flow into a modal and being able to trigger it from anywhere soon IMO. Wdyt?

### DIFF
--- a/src/lib/components/stepper/stepper.svelte
+++ b/src/lib/components/stepper/stepper.svelte
@@ -10,7 +10,7 @@
   import modal from '$lib/stores/modal';
   import { browser } from '$app/environment';
 
-  const dispatch = createEventDispatcher<{ stepChange: never; conclude: never }>();
+  const dispatch = createEventDispatcher<{ stepChange: never }>();
 
   export let steps: Steps;
   export let context: (() => Writable<unknown>) | undefined = undefined;
@@ -198,8 +198,6 @@
       disableTransitions = false;
     } else {
       modal.hide();
-      // dispatch in case Stepper is being used in a standalone page (like /funder-onboarding)
-      dispatch('conclude');
     }
   }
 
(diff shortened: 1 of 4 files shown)


evvvritt:
Shouldn't there be _some_ Back button in the flow?

Even when we make the modal version, it should have a back button? And I expect we'll still have a dedicated route to share directly with funders when we manually outreach?

(also, this reverts the target="_blank" links on the homepage which I don't see why are there?)

efstajas:
> Shouldn't there be _some_ Back button in the flow? Even when we make the modal version, it should have a back button?

Once it's in a modal, IMO it should have "continue" and "cancel" buttons, like other flows-in-modals already do. IIRC there's also a discussion w/ Brandon to always have "Cancel" buttons visible in flows, so it could be improved together with that.

>  And I expect we'll still have a dedicated route to share directly with funders when we manually outreach?

Yeah, I think we can keep the dedicated route still, it'd be easy to import the flow and put it in the dedicated route after the refactor. IMO the "dedicated version" of this flow shouldn't have a back or cancel button, because it's just a page that you're being linked to directly, so your cancel button is just closing the tab.

> (also, this reverts the target="_blank" links on the homepage which I don't see why are there?)

I did that initially _because_ there's no back button on the dedicated pages, and so I felt it more appropriate to create a new context by opening these flows in another tab. In my head we'd always put these flows into a modal eventually, at which point having a back / cancel button is trivial (you just close the modal). Don't feel strongly about this though!

evvvritt (approved):
oki

PR #685, opened 2023-09-25
feature branch: multi drip lists

efstajas:
https://www.figma.com/file/vyI7f996JF8zwhnXwAwXdC/%F0%9F%92%A7-Drips?type=design&node-id=3938%3A9644&mode=design&t=iIZX0ox2PnvqK1Vb-1


efstajas:
@evvvritt any idea what broke tooltips on lists in this PR? 🤔 

Screenshot 2023-10-19 at 16 29 32

evvvritt:
Not immediately no. Do we need Tooltips there? Could we just directly link to profiles/project/drip-list pages?

efstajas:
It's already linking directly, but I think tooltips are quite useful, yeah. It's the only way you can see the full address of a recipient for example, and copy it without having to navigate to the profile and back. Also allows you to quickly see who's in control of the funds going to a claimed project.

efstajas:
@evvvritt fixed it — problem was that the `transform` on the drip-list-card established a new positioning context, but the tooltips rely on being positioned relative to the viewport.

efstajas:
Todo: remove the half-implementation of #698 (new metadata v3). It's halfway done here, but we're gonna launch this later.

PR #691, opened 2023-09-26
feat: simplified collect screen

efstajas:
Most users will only be collecting split income (from Drip Lists and Projects), so this aims to simplify their collect screen by hiding most streaming-related stuff.

- Completely hides streaming-related balances if they're zero
- Hides squeezing if they're zero
- Greatly simplifies language around collectable balances — no more "cycle", no more "splittable", etc.… just "settlement once per week".

Also:
- Fix for `AnnotationBox` having a wonky layout if actions slot is empty (cc @evvvritt)

Someone who is collecting income from Drip Lists / Projects & also has some stream income now sees the screen like this: 

Screenshot 2023-10-02 at 17 14 30

### DIFF
--- a/src/e2e-tests/top-up-create-stream.e2e.test.ts
+++ b/src/e2e-tests/top-up-create-stream.e2e.test.ts
@@ -387,7 +387,7 @@ describe('app', async () => {
       });
 
       it('expands the squeezing section', async () => {
-        await page.locator('label:has-text("Include funds from current cycle")').click();
+        await page.locator('label:has-text("Include funds streamed in current cycle")').click();
 
         await page
           .locator(`data-testid=item-383620263794848526656662033323214000554911775452`)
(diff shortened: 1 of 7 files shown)


evvvritt:
Looks good except I'd remove the "+" next to the earned amount in this simplified version. (Perhaps other contexts it still makes sense?)

evvvritt (changes requested):
(remove the "+" next to the Earned amount in the simplified version)

efstajas:
> Looks good except I'd remove the "+" next to the earned amount in this simplified version. (Perhaps other contexts it still makes sense?)

good point, one sec!

brandonhaslegs:
- Should "Streams" be capitalized like Projects and Drip Lists?
- Instead of "on a weekly cycle" use "on a weekly basis"
- Can you actually only collect on a weekly cycle? Like if you miss the date, you have to wait another week to collect? Or is money **_added_** on a weekly cycle but you can **_collect_** at any time (after the first cycle)? Should make that more clear.
- Is there a situation where the two amounts would be different?
- In the first row, will the text "From Project or Drip Lists" change if funds have also come from Streams?

brandonhaslegs (approved):

evvvritt (approved):

PR #692, opened 2023-09-26
style: drip-list-card has full 'copy link' button (closes #625)

### LINKED ISSUE #625: Edit the copy link button
looks weird in that light grey and it's not totally clear what it is. Text helps. make it a ghost button to push the hierarchy down a bit.

image

evvvritt:
closes #625


efstajas:
I think would be cool if you could make the `ShareButton` actually have the label? The share button actually brings up the native share API on mobile / other platforms where it's a thing, and becomes a "copy link" button only if no share dialog is available.

Label could be `Share` if `shareSupported` is `true`, and `Copy link` if `false`.

efstajas (approved):
nice!

PR #705, opened 2023-10-03
feat: drip list support button

efstajas:
Screenshot 2023-10-03 at 15 07 31

### DIFF
--- a/src/lib/components/support-card/support-card.svelte
+++ b/src/lib/components/support-card/support-card.svelte
@@ -20,15 +20,40 @@
   import Stepper from '$lib/components/stepper/stepper.svelte';
   import editDripListSteps from '$lib/flows/edit-drip-list/edit-drip-list-steps';
   import buildUrl from '$lib/utils/build-url';
+  import type { SplitsEntry } from 'radicle-drips';
 
-  export let project: ClaimedGitProject;
+  export let project: ClaimedGitProject | undefined = undefined;
+  export let dripList: DripList | undefined = undefined;
 
-  let dripList: DripList | null | undefined = undefined;
-  let isSupportingProject: boolean | undefined = undefined;
+  let ownDripList: DripList | null | undefined = undefined;
+  let ownDripListSplits: SplitsEntry[] | undefined = undefined;
+
+  let isSupporting: boolean | undefined;
+  $: isSupporting = ownDripListSplits?.some(
+    (s) =>
+      s.accountId === project?.repoDriverAccount.accountId ||
+      s.accountId === dripList?.account.accountId,
+  );
+
+  $: isOwner =
+    $walletStore.dripsAccountId === project?.owner.accountId ||
+    $walletStore.dripsAccountId === dripList?.account.owner.accountId;
+
+  let supportUrl: string;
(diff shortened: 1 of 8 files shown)


src/lib/components/support-card/support-card.svelte
```diff
@@ -88,35 +134,28 @@
...
   <p>
-    {#if isSupportingProject}
-      You're already supporting <span class="typo-text-bold">{project.source.repoName}</span> with your
-      Drip List.
+    {#if isSupporting}
+      You're already supporting this with your Drip List.
     {:else}
-      Add <span class="typo-text-bold">{project.source.repoName}</span> to your Drip List to support
-      this project.
+      Add this {project ? 'project' : ''} to your Drip List to flexibly support it with an ongoing contribution.
```
evvvritt, line 149:
```suggestion
      Add this {project ? 'project' : ''} to {ownDripList === null ? 'a' : 'your'} Drip List where they will receive a portion of any funds streamed to your list.
```

src/lib/components/support-card/support-card.svelte
```diff
@@ -88,35 +134,28 @@
...
     {/if}
   </p>
   <Button
     on:click={handleSupportButton}
-    disabled={isSupportingProject}
+    disabled={isSupporting || isOwner}
     size="large"
     icon={Plus}
-    variant="primary">{dripList === null ? 'Create your Drip List' : 'Add to Drip List'}</Button
+    variant="primary">{ownDripList === null ? 'Create your Drip List' : 'Add to Drip List'}</Button
```
evvvritt, line 157: maybe nice if they're already supporting the button is disabled with checkmark icon and "Added to your Drip List"?

evvvritt (approved):
Approved with minor copy suggestions

src/lib/components/support-card/support-card.svelte
efstajas: yeah tru

efstajas: though irrelevant very soon when we launch the multi drip list feature, so i'd keep it like this for now!

src/lib/components/support-card/support-card.svelte
brandonhaslegs: I think it sounds weird to use "where" here

evvvritt: "**and** they will.."
would also do

brandonhaslegs: Yeah that's better. But maybe "and _it_ will" since it's referring to the list.

evvvritt: It's referring to the project though (which receives a portion of what goes to the list). 'Become a supporter of the project by adding it to a Drip List'

evvvritt: Ah, right this is used on project and drip list pages* So ya maybe a conditional for they/it

brandonhaslegs: It works for project and list. Just because it's a project doesn't mean it's multiple people.

PR #706, opened 2023-10-03
style: use drip-list-badge in streams table (closes #700)

### LINKED ISSUE #700: Use the Drip List name instead of “Your Drip List” as stream recipient for Support Streams
Screenshot 2023-09-27 at 16 17 05

The stream recipient in streams tables globally should always use the Drip List badge component rather than a static "Your Drip List" name.

evvvritt:
closes #700 (don't see a reason not to just merge this into main?)

(would've been nice to hide the drip list username when in dashboard streams context but seems really hairy to pass that context down to the UserBadgeCell component?)

### DIFF
--- a/src/lib/components/drip-list-badge/drip-list-badge.svelte
+++ b/src/lib/components/drip-list-badge/drip-list-badge.svelte
@@ -4,7 +4,7 @@
   import formatAddress from '$lib/utils/format-address';
 
   export let listId: string;
-  export let listName: string | undefined = undefined;
+  export let listName: string | '##loading##' | undefined = undefined;
   export let owner: string | undefined = undefined;
 
   export let isLinked = true;
@@ -41,7 +41,8 @@
     <div class="name typo-text text-foreground flex-1 min-w-0 truncate">
       <span
         >{#if username}<span class="text-foreground-level-5">{username}/</span
-          >{/if}{@html listName}</span
+          >{/if}{#if listName === '##loading##'}<span class="animate-pulse">...</span
+          >{:else}{listName}{/if}</span
       >
     </div>
   {/if}
--- a/src/lib/components/drip-list-badge/drip-list-badge.svelte
+++ b/src/lib/components/drip-list-badge/drip-list-badge.svelte
@@ -3,53 +3,51 @@
   import ensStore from '$lib/stores/ens';
   import formatAddress from '$lib/utils/format-address';
 
-  export let listName: string;
   export let listId: string;
-  export let owner: string;
+  export let listName: string | undefined = undefined;
+  export let owner: string | undefined = undefined;
 
-  export let showAvatar = true;
-  export let showName = true;
   export let isLinked = true;
+  export let showAvatar = true;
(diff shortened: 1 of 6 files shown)


efstajas (changes requested):
Would be great to make a small version of the drip list badge icon to align it with address stream recipients

Screenshot 2023-10-04 at 12 26 01

Also re. showing the ens / address in the drip list badge, I guess streams always appear in the context of one particular user, so I feel like we could just always hide it in the context of the stream table. Then it'd need to be one boolean prop like `showUsername` 🤔

evvvritt:
Ok, simplified things, lmk wyt @efstajas 
- owner + listName are optional (are hidden if not provided)
- animated loading handled by listName input
- small icon size version

src/lib/components/drip-list-badge/drip-list-badge.svelte
```diff
@@ -3,53 +3,51 @@
...
+      />
     </div>
   {/if}
-  {#if showName}
+  {#if listName}
     <div class="name typo-text text-foreground flex-1 min-w-0 truncate">
-      <span><span class="text-foreground-level-5">{username}/</span>{listName}</span>
+      <span
+        >{#if username}<span class="text-foreground-level-5">{username}/</span
+          >{/if}{@html listName}</span
```
efstajas, line 44: This seems to open up a XSS vulnerability… could call my list `<script>doBadStuff()</script>` 🥴 

maybe `listName` could be `string | null | undefined` with a default of `null`, where `null` means "no name", and `undefined` means "name is loading"

efstajas (changes requested):

src/lib/components/drip-list-badge/drip-list-badge.svelte
evvvritt: oddly, svelte transforms undefined to null inside the component!
```
listName={dripList ? dripList.name : undefined}

```
outputs null inside the component...
maybe they consider it an error?

efstajas: ... but tbh, maybe just a global "loading" prop is way easier and more explicit than any of that 😄

src/lib/components/drip-list-badge/drip-list-badge.svelte
```diff
@@ -3,53 +3,52 @@
   import ensStore from '$lib/stores/ens';
   import formatAddress from '$lib/utils/format-address';
 
-  export let listName: string;
   export let listId: string;
-  export let owner: string;
+  export let listName: string | '##loading##' | undefined = undefined;
```
efstajas, line 7: not a _huge_ fan of using a magic literal value with a user-defined string, but whatever, quite unlikely anyone will call their list this^^

efstajas (approved):

src/lib/components/drip-list-badge/drip-list-badge.svelte
evvvritt: oki, added a listLoading flag instead

PR #710, opened 2023-10-04
load optimized Inter fonts (closes #709)

### LINKED ISSUE #709: Optimize asset loading (duplicate Inter fonts, ...)
For some reason we're loading two version of Inter regular as an otf? Is it available as a .woff which is maybe optimized / smaller file size? (or we can subset it to the glyphs we need)
(Inter Regular x2 + Semibold accounts for 0.75MB of pagesize 😅)

assert.js is also rather large?

evvvritt:
loads local .woff2 files (overwrites `radicle-design-system/typography.css` `@font-face` rules)

alternatively, we could update the `radicle-design-system` css + font files but I don't have write access so thought this was faster

tested on deploy and it properly loads the .woff2 files instead of the .otf (locally it sometimes loads the .otf)


efstajas (changes requested):
Nice, loads a lot faster! Check this section on the LP tho:

Screenshot 2023-10-04 at 16 32 20

The numbers wiggle around, so it seems like the `tabular nums` font feature stopped working?

evvvritt:
Ah true, it looks like I can preserve them in the optimizer from font-squirrel (options below). 

Though I'm having trouble identifying some of the other font features we have in our css: (@brandonhaslegs dyk?)
Screenshot 2023-10-04 at 18 39 44

1. I presume `ss01`, `ss02` are "Styled Set 1, ...2" in the options below?
2. what is `cv01`?
3. what is `calt`?

Screenshot 2023-10-04 at 18 39 36

efstajas:
You can see them all here: https://rsms.me/inter/lab/

brandonhaslegs:
- CV01 is the alternate `1` character. It's not necessary and I don't know why it was turned on.
- turning CALT off makes the 0x1 issue go away. If it's on then the `x` character turns into a `×` character. This one is important.

evvvritt:
Ok thanks, of course when I enable tabular numbers in the generator it breaks 🤣

evvvritt:
It's a "known bug" that the font files aren't included in expert mode - but realistically no fix in sight. @efstajas / @brandonhaslegs do you guys know other ways to optimize the .otf's into .woff with our font features enabled?

brandonhaslegs:
@evvvritt if you have the optimized OTF's, then you can just convert them to WOFF's on cloudconvert.com.

efstajas (approved):

Issue #704, opened 2023-10-02
Add live Drip List on landing page


brandonhaslegs:
Find the design in Figma [here](https://www.figma.com/file/vyI7f996JF8zwhnXwAwXdC/%F0%9F%92%A7-Drips?type=design&node-id=4350%3A18134&mode=design&t=Gro7UJcp3br2aLBO-1).

<img width="1154" alt="image" src="https://github.com/drips-network/app/assets/4406983/b03ddc29-00bc-4494-8241-67b6578e3f7c">

PR #725, opened 2023-10-09
Add 'Total Dripped' badge to landing page (closes #645)

### LINKED ISSUE #645: Add Real-Time $ Value of All Tokens on the Protocol
To be clear, this is the total amount of funds that are in Drips. Not the total amount that has been streamed. The sum of everyone's top ups.

image

evvvritt:
for #645

I've started with the discussed idea to (for now) just use the 'total dripped' amount to Radworks Software Dependencies list, since it's an increasing number.

I copied logic from `drip-list-card.svelte` but the balances/streams to the list aren't being loaded.

@efstajas, does this have to do with this logic being in the homepage, when the store logic is still under /app?
Guidance on how to initialize / lookup inside the landing page?
(we'll need this for featuring that list lower on the landing page #704 to right?)

### DIFF
--- a/src/lib/components/network-total-dripped-badge/NetworkTotalDrippedBadge.svelte
+++ b/src/lib/components/network-total-dripped-badge/NetworkTotalDrippedBadge.svelte
@@ -1,0 +1,48 @@
+<script lang="ts">
+  import Drip from '../illustrations/drip.svelte';
+  import mergeAmounts from '$lib/utils/amounts/merge-amounts';
+  import balancesStore from '$lib/stores/balances/balances.store';
+  import { constants } from 'radicle-drips';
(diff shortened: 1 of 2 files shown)


evvvritt:
Ok, trivially tried to simply load the Radworks drip list on the loading page. Came close but still couldn't figure out how to load balances / fiat estimator on landing page. 

I'm gonna put this aside until @efstajas is back because maybe we want to rethink how we make this type of data/components workable on landing page in general—right now they need to be under /app route.

Screenshot 2023-10-11 at 17 18 41

evvvritt:
As discussed on the call today, next attempt at badge should:
- load fiat estimate store on landing page
- use static values for USDC and RAD from the radworks drip list
- design: "$ XXX dripped" animated going up

PR #726, opened 2023-10-09
style: rmv 'funds' in Add modal (closes #719)

### LINKED ISSUE #719: Update "Add USDC funds" Modal
Remove the word "Funds" from this title.

(We have "Collect USDC" and "Withdraw USDC" but "Add USDC" is the only one with "funds" in the title.)

image

evvvritt:
closes #719

### DIFF
--- a/src/lib/flows/top-up-flow/enter-amount.svelte
+++ b/src/lib/flows/top-up-flow/enter-amount.svelte
@@ -100,7 +100,7 @@
 <StepLayout>
   <EmojiAndToken emoji="💰" tokenAddress={tokenInfo.info.address} animateTokenOnMount />
   <StepHeader
-    headline={`Add ${tokenInfo?.info.symbol ?? ''} funds`}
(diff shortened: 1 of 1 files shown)


src/lib/flows/top-up-flow/enter-amount.svelte
```diff
@@ -100,7 +100,7 @@
 <StepLayout>
   <EmojiAndToken emoji="💰" tokenAddress={tokenInfo.info.address} animateTokenOnMount />
   <StepHeader
-    headline={`Add ${tokenInfo?.info.symbol ?? ''} funds`}
+    headline={`Add ${tokenInfo?.info.symbol}`}
```
jtourkos, line 103: Is there a reason we had a default `''` value before?

evvvritt: Unclear to me, but the changes match what we have in Withdraw modal (no ' ' fallback)

jtourkos (approved):

PR #743, opened 2023-10-18
feat: #739

### LINKED ISSUE #739: Allow adding a Drip List to project dependencies
You can already add external Drip Lists to your Drip Lists, but you can't add Drip Lists to a project's dependencies yet. It'd be pretty cool for composability if that were possible, and would also make it easier to both receive funds for your dependencies through your project, and at the same time also support them yourself (through the Drip List).

efstajas:
Resolves #739

### DIFF
--- a/src/lib/utils/project/GitProjectService.ts
+++ b/src/lib/utils/project/GitProjectService.ts
@@ -462,7 +462,7 @@ export default class GitProjectService {
 
         dependenciesSplitMetadata.push(receiver);
         receivers.push(receiver);
-      } else {
+      } else if (isValidGitUrl(urlOrAddress)) {
         const { forge, username, repoName } = GitProjectService.deconstructUrl(urlOrAddress);
 
         const receiver = {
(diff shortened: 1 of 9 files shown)


src/lib/utils/project/GitProjectService.ts
```diff
@@ -462,7 +462,7 @@ export default class GitProjectService {
 
         dependenciesSplitMetadata.push(receiver);
         receivers.push(receiver);
-      } else {
+      } else if (isValidGitUrl(urlOrAddress)) {
```
jtourkos, line 465: Maybe we should rename `urlOrAddress` to something else now we support Drip Lists?
Something more generic like `dependencyIdentifier` or similar?

src/lib/utils/project/GitProjectService.ts
```diff
@@ -661,26 +671,82 @@ export default class GitProjectService {
...
-            ? {
-                type: 'repo',
-                weight: d.weight,
-                account: {
-                  driver: 'repo',
-                  accountId: d.accountId,
-                },
-                source: d.source,
+        dependencies: await Promise.all(
+          projectMetadata.data.splits.dependencies.map(async (d) => {
```
jtourkos, line 675: Not for now and it's a personal preference, but since this might be changing in the future I'd abstract the whole dependencies calculation to something else.
Probably a separate function for now, and I'd make the comments functions themselves.

jtourkos (approved):

src/lib/utils/project/GitProjectService.ts
efstajas: yeah

src/lib/utils/project/GitProjectService.ts
efstajas: Yeah, I also had an urge to refactor here, but decided not to given all this will be moot once we have the new service

PR #753, opened 2023-10-19
feat: stable drip lists view

efstajas:
Makes the drip list "thumb" card layout stable while loading, and neatly interpolates the height of the splits on the single drip list view:

https://github.com/drips-network/app/assets/1018218/ff92b52b-e5f4-4788-bcd3-186acd4bf9dc

Caveat is that this no longer shows list descriptions in listing contexts… Which I think is probably fine, since they still exist when viewing a single drip list directly. If we definitely want to show them, we could:

1. Show max a single truncated line of description, in a container exactly as high as one row in the splits component 
2. reduce the max rows in the split component for drip lists card w/ description by 1

… then the height would still remain stable. Wdyt?

### DIFF
--- a/src/lib/components/drip-list-card/drip-list-card-representational.svelte
+++ b/src/lib/components/drip-list-card/drip-list-card-representational.svelte
@@ -234,7 +234,6 @@
     top: 0;
     right: 0;
     bottom: 0;
-    /* gradient from transparent to var(--color-background), left to right */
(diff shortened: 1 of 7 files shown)


brandonhaslegs:
Do 1, I think the descriptions are important.

efstajas:
@brandonhaslegs Alright, plan worked out 👍 

https://github.com/drips-network/app/assets/1018218/f78b89bf-03da-45d8-b157-12acce294dfc

PR #771, opened 2023-10-27
chore: upgrade to svelte v4

efstajas:


efstajas:
Closing this for now, again. Once we moved all stream & balance estimation logic to the new GraphQL API, we'll be able to keep the props for the table stable, and only re-render the amount components locally rather than the entire table. That'll greatly increase performance, and also get rid of a blocker for merging this, which is that with Svelte V4 for some reason a lot more DOM nodes are being rapidly replaced within the table compared to V3.

PR #774, opened 2023-11-01
remove beta badge from headers

evvvritt:
no need to wait for other landing page updates right?


brandonhaslegs:
> no need to wait for other landing page updates right?

Do we?

efstajas (approved):

PR #802, opened 2023-11-13
fix: show funding.json mismatch error to user

evvvritt:
(instead of "not found" for any error, including mismatched content)

### DIFF
--- a/src/routes/app/(flows)/claim-project/steps/add-ethereum-address/add-ethereum-address.svelte
+++ b/src/routes/app/(flows)/claim-project/steps/add-ethereum-address/add-ethereum-address.svelte
@@ -48,24 +48,20 @@
 
         const { forge, username, repoName } = GitProjectService.deconstructUrl($context.gitUrl);
 
-        try {
-          await github.getFundingJson(
-            username,
-            repoName,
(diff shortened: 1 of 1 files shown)


src/routes/app/(flows)/claim-project/steps/add-ethereum-address/add-ethereum-address.svelte
```diff
@@ -48,24 +48,20 @@
...
-            dripsJsonTemplate(
-              $walletStore.address ?? unreachable(),
-              $walletStore.network.name
-                ? $walletStore.network.name === 'homestead'
-                  ? 'ethereum'
-                  : $walletStore.network.name
-                : unreachable(),
-            ),
-          );
+        await github.getFundingJson(
```
jtourkos, line 51: What'll happen now on error?
Is it caught somewhere higher on the call stack?

evvvritt: Yes the Stepper catches any error and displays it in the same error modal style

efstajas (approved):

PR #851, opened 2023-12-06
chore: pump dependencies

jtourkos:


efstajas:
@jtourkos could you check the errors?

PR #863, opened 2023-12-11
rename streams -> funds

### LINKED ISSUE #813: Rename "Streams" to "Funds" in the nav
image

efstajas:
Chose this icon because we use it for the Balances section, but we should probably make an actual "tokens" icon?

Resolves #813


evvvritt (approved):
ehh Orgs icon looks like people and weird there. Let's use the Token Streams icon and Coin for Balances (added commits) – but ya would be nice to have a better balances icon ⚖️

PR #884, opened 2024-01-04
style: update spinner to falling drip animation (closes #815)

### LINKED ISSUE #815: Update loading animation
Update it to an SVG of the droplet icon
image

and I'm thinking you just animate that icon similar to the moving droplets on the home page. So they start out tiny at the top of the loader (and invisible) then it gets big quickly as it moves down and then gets small as it decreases in size again.

evvvritt:
- implements @brandonhaslegs's first design in #815 which I prefer :}
- fully scraps the old Spinner.svelte code (lmk if we should keep?)
- appears next to logo on certain loads (like project profiles)

click/tx around and see!


brandonhaslegs:
The animation is showing up above the globe icon and inside the connect button.

https://github.com/drips-network/app/assets/4406983/de5a3c25-8c6d-4fda-b3bc-2147270d9aa7

efstajas:
@brandonhaslegs both of these are technically loading for a very short while. The spinner right now already appears in these two places, it's just less visible because it's grey. 

We could tweak the spinner to only appear after a delay of 200ms or so, then it wouldn't look so glitchy in these kinds of cases.

brandonhaslegs:
Yeah just whatever you can do to make this not happen. Why does it have to re-load the asset every time? Can't we cache some of these things?

evvvritt:
added a 300ms delay if y'all can re-review :)

brandonhaslegs (approved):

brandonhaslegs:
Do 2 too! can't be that hard to change the bezier

evvvritt:
ok, did `#1`
for `#2` I've set easing to `ease`
if you're not happy with it i made this [jsfiddle](https://jsfiddle.net/bz7drfyu/52/) for you guys to play with alternative easings/bezier (change the dropdown or paste a custom bezier function from something like [easings.co](https://easings.co/))

efstajas (approved):
LGTM with `ease` 👍

PR #885, opened 2024-01-04
fix: modal should scroll when mouse above overlay (closes #859)

### LINKED ISSUE #859: Modals should be scrollable when the cursor is outside the modal
Modals are currently only scrollable in the area marked in red. They should also be scrollable when you scroll on the area marked in yellow.

Screenshot 2023-12-10 at 12 17 51

evvvritt:
fixes #859


efstajas (approved):
lgtm 👍

PR #886, opened 2024-01-04
refactor: open create-drip-list and claim-project as modals (closes #731)

### LINKED ISSUE #731: Create Drip List and project claim modals
Separate the modal into a floating modal that is accessible from the app and doesn't go to a full page screen. Not sure if there's an issue for this, but we've talked about it a lot.

And of course, keep the single page version for when someone is linked to it from an external link.

evvvritt:
closes #731


evvvritt:
@efstajas can you take a look at this draft and let me know if this is the correct approach and I'll apply for claim-project?

basically migrated the funder-onboarding stepper into it's own folder `create-drip-list-flow` with a new `CreateDripListStepper.svelte` component that's used in the `/app/funder-onboarding/+page.svelte` route

efstajas:
yep that sounds about right! that would've been my approach too - make it a normal flow in the `flows` folder, and then embed that back into the dedicated routes.

efstajas:
Re 1: Ah true, that's tricky but should be possible with sidestep. The general way to do it would be to make the customize modal a step (as in wrap it in a step layout etc.) and then when the user clicks "Customize", emit a `sidestep` event with a `steps` value that is just an array with the single step in it. When the user clicks confirm on the sidestep, emit `conclude`. I can also give this a try if you want.

Re 2: Check out the `condition` field that you can set for steps: https://github.com/radicle-dev/drips-app-v2/blob/4ada9c21582703a2b89812d8908bc7111cfa4adc/src/routes/app/(flows)/funder-onboarding/funder-onboarding-flow.ts#L136

I think after adding this, I only refactored the funder onboarding flow to use it, not the project claim flow. Basically, condition is a function that returns a boolean, and is called whenever the stepper wants to navigate to the step. If the value returned is false, it gets ignored completely, as if it was never part of the steps array. I would try maybe adding a `subgraphPolledSuccess` boolean to the step state that is flipped to `true` when `PollSubgraph` finishes. And then skip `PollSubgraph` if this boolean is true. That should result in PollSubgraph being skipped when navigating back from subsequent steps.

evvvritt:
Ok, should make @brandonhaslegs very happy 🤣 

Encourage the following tests while reviewing:
1. create drip list from /app/drip-lists
2. create drip list from /app/funder-onboarding
3. claim project from /app/projects
4. claim project from /app/claim-project

@efstajas fyi, I slightly increased testing timeout to 7s because they fail often on github?

src/lib/components/projects-section/projects-section.svelte
```diff
@@ -65,10 +66,11 @@
       ? [
           {
             // TODO: (FIX) clicking this button after completing the claim project flow freezes the UI (in all browsers). It shouldnʼt.  😊
-            handler: () => goto(`/app/claim-project`),
+            // handler: () => goto(`/app/claim-project`),
```
efstajas, line 69:
```suggestion
```

efstajas (approved):
Just one leftover comment, please accept the suggestion before merge. Otherwise LGTM!! 😎

PR #900, opened 2024-01-12
Custom project logos / avatars

efstajas:


efstajas:
@evvvritt checks fail because the new GQL types for this aren't on prod yet, but they will be soon, then I'll re-run the checks. You can already take a look at the code :)

efstajas:
Do not merge until both infra slots are deployed with the new API changes. Currently only slot #1 has the changes deployed.

evvvritt (changes requested):
Works well but think it should update or refresh the project profile page on tx confirm 🤓

evvvritt (approved):

PR #918, opened 2024-01-22
overhaul e2e setup

efstajas:
The tests should work a lot better and more consistently all around now, and not slow us down anymore. The main improvements are:

- If you just run npm run e2e, the tests now build & execute within Docker, meaning there will no longer be inconsistencies due to local environment differences. Of course, you can still run E2E tests in headless mode locally (see DEVELOPMENT.md below). While the app is built within the image at runtime, it's smart enough to only run npm ci when package.json has been changed. Otherwise, it'll just copy over your latest local source code into the container, and build it.
- The Docker build cache for the services (events-processor, gql-api etc.) is now automatically invalidated when a new commit has been made on the respective repo's main branch. This means they're effectively always up to date now. No need to manually delete and rebuild images anymore.
- You can specify a specific branch to build events-processor and gql-api from by simply setting an environment variable (see DEVELOPMENT.md below). This means it's now super easy to test changes to API / Event processor in the context of an entire local Drips stack.
- Simplified local-env commands for running the app locally connected to the mock environment.


efstajas:
svelte-check fails bc infra slot #1 has the latest changes for the custom avatar API deployed for testing. This PR is already ready for review, and tests should pass after merging #900 and updating this branch from main.

evvvritt (approved):

PR #920, opened 2024-01-23
profile: empty sections ordered last in stack (closes #723)

### LINKED ISSUE #723: Profile: move active sections to the top
Nicer to see the drip list than the empty projects section ya?
hide or put empty sections down the page?
Screenshot 2023-10-09 at 13 55 57

evvvritt:
closes #723


efstajas (changes requested):
It made the layout super unstable. Since the sections resolve at very different times now, they're just jumping around for quite a long time after they first appear. Try loading this page without cache for instance: https://deploy-preview-920--drips-app-v2.netlify.app/app/breadblush.eth

If we're gonna do dynamic ordering I think we would have to show the entire layout / sections only once they've all loaded. Otherwise the layout will always be unstable.

Alternatively, what if we just collapse empty sections by default? It'd still be a bit wonky, but at least the wonkiness would be transitioned nicely.

PR #926, opened 2024-01-30
Round split receiver weights to reach 100%

jtourkos:


src/lib/utils/round-weights.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      if (roundingRemainder > 0) {
+        additionalWeight = 1;
+        roundingRemainder -= 1;
+      }
+      return {
+        ...r,
+        weight: equalWeight + additionalWeight,
+      };
+    });
```
efstajas, line 40: This would probably never happen, but what if there's only a very few receivers with equal weights and the remainder is higher than the number of receivers? 🤔

efstajas (approved):

jtourkos:
@efstajas should we merge this?

PR #940, opened 2024-02-07
fix: rounding logic

jtourkos:
@efstajas here's a fix for the rounding logic. There were a few edge cases that this PR should fix.
However, I still haven't found the cause of the receivers mismatch (onchain vs metadata) even though I spotted a wrong related assignment.

If we are unsure about this, we can also revert all rounding.

Regarding the reported issue on [Discord](https://discord.com/channels/1096003917717983252/1204581052962578512) we need to suggest the following configuration:
```json
         {
            "accountId": "246143406710477020717494096468966982681784893038",
            "weight": "187500"
          },
          {
            "accountId": "1208577989352973003091899595206122026695351403583",
            "weight": "562501" 👈
          },
          {
            "accountId": "80924457310386880091302880843407861917397762025481227787685752471552",
            "weight": "83333"
          },
          {
            "accountId": "80928961545538538466165477808532767810848587485712440198768076259328",
            "weight": "83333"
          },
          {
            "accountId": "81036473733089919926028385797282177451059549353244988582777280502540",
            "weight": "83333"
          }
```

### DIFF
--- a/src/lib/utils/project/GitProjectService.ts
+++ b/src/lib/utils/project/GitProjectService.ts
@@ -30,6 +30,7 @@ import { Driver, Forge } from '$lib/graphql/__generated__/base-types';
 import GitHub from '../github/GitHub';
 import { Octokit } from '@octokit/rest';
 import roundWeights from '../round-weights';
+import unreachable from '../unreachable';
 
 export default class GitProjectService {
   private _github!: GitHub;
@@ -328,8 +329,6 @@ export default class GitProjectService {
     maintainerListEditorConfig: ListEditorConfig,
     dependencyListEditorConfig: ListEditorConfig,
   ) {
-    const receivers: SplitsReceiverStruct[] = [];
-
     // Populate dependencies splits and metadata.
     const dependenciesInput = Object.entries(dependencyListEditorConfig.percentages);
 
@@ -354,7 +353,6 @@ export default class GitProjectService {
         };
 
         dependenciesSplitMetadata.push(receiver);
-        receivers.push(receiver);
       } else if (isValidGitUrl(itemId)) {
         const { forge, username, repoName } = GitProjectService.deconstructUrl(itemId);
 
@@ -373,7 +371,6 @@ export default class GitProjectService {
           ...receiver,
           source: GitProjectService.populateSource(forge, repoName, username),
         });
-        receivers.push(receiver);
       } else {
         // It's the account ID for another Drip List
         const receiver = {
@@ -383,7 +380,6 @@ export default class GitProjectService {
         };
 
         dependenciesSplitMetadata.push(receiver);
-        receivers.push(receiver);
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/project/GitProjectService.ts
```diff
@@ -408,17 +404,51 @@ export default class GitProjectService {
...
+        ...d,
+        typeOfReceiver: 'Dependency',
+      }))
+      .concat(
+        maintainersSplitsMetadata.map((m) => ({
+          ...m,
+          typeOfReceiver: 'Maintainer',
+        })),
+      );
+    const roundedReceivers = roundWeights(dependenciesAndMaintainers as any);
```
efstajas, line 420: why do we need `as any` here?

src/lib/utils/project/GitProjectService.ts
```diff
@@ -408,17 +404,51 @@ export default class GitProjectService {
...
+      .filter((r) => r.typeOfReceiver === 'Maintainer')
+      .map((m) => {
+        const maintainer =
+          maintainersSplitsMetadata.find((s) => s.accountId === m.accountId) ?? unreachable();
+
+        return {
+          ...maintainer,
+          weight: m.weight,
+        };
+      });
```
efstajas, line 442: nitpick, but could extract this map fn and reuse it for both of these

src/lib/utils/project/GitProjectService.ts
```diff
@@ -408,17 +404,51 @@ export default class GitProjectService {
...
+        typeOfReceiver: 'Dependency',
+      }))
+      .concat(
+        maintainersSplitsMetadata.map((m) => ({
+          ...m,
+          typeOfReceiver: 'Maintainer',
+        })),
+      );
+    const roundedReceivers = roundWeights(dependenciesAndMaintainers as any);
+    const roundedDependencies = roundedReceivers
```
efstajas, line 421: let's call these "dependenciesMetadataRounded" and "maintainersMetadataRounded"? just a bit more explicit & easier to understand IMO

src/lib/utils/round-weights.ts
```diff
@@ -1,4 +1,11 @@
-import { constants, type SplitsReceiverStruct } from 'radicle-drips';
+import { BigNumber } from 'ethers';
+import { constants } from 'radicle-drips';
+
+export interface TypedReceiver {
+  typeOfReceiver: 'Dependency' | 'Maintainer';
+  accountId: string;
+  weight: number;
+}
```
efstajas, line 8: again a bit of a nitpick, but if you type `roundWeights` as `roundWeights<T extends { accountId: string, weight: number }>(receivers: T[]): T[]`, you could pass any object in that has at least accountId and weight, and the return type would still include any additional fields (e.g. `typeOfReceiver` here).

would just make it a bit more generic 🤷

efstajas (changes requested):

PR #942, opened 2024-02-08
refactor: always apply gas buffers when needed

efstajas:
The `transact` utility now expects an array of `PopulatedTransactions`, each with an `applyGasBuffer` property. If at least one of the transactions passed has this set to true, `transact` will automatically simulate the entire array of transactions on Tenderly, in sequence, and then apply a 15% gas buffer to the resulting gas estimation before it proposes to the wallet.

Batch transactions proposed to a Safe will still always be estimated on Tenderly. 

All other changes are refactors to make everything pass an array of PopulatedTransactions to `transact`. Most of them are minor, expect `DripListService`, where a few bigger changes were needed. It's a lot simpler now though!


jtourkos (approved):
Nice! 👍

PR #948, opened 2024-02-09
explore page 

efstajas:
adds
- explore page at /app
- new developer section for stream detail view


efstajas:
failed tests are due to API not being live in slot 1 yet

evvvritt (approved):

PR #990, opened 2024-02-20
restore drips logo on mobile when disconnected (closes #962)

### LINKED ISSUE #962: Show Drips Logo on mobile
at least when user's aren't connected. we could move the "Connect" button to the bottom edge of the screen, fixed? (wdyt @brandonhaslegs) then the menu would take it's place once connected.

Screenshot 2024-02-13 at 20 01 34

evvvritt:
ALSO, moved Settings button to inside the Account Menu when wallet-connected (more space)


brandonhaslegs:
logo gets bigger in the app, should be identical to the landing page.

evvvritt:
(todo: remove "Code" from header as discussed in call)

efstajas (changes requested):
looks like this for me:

Screenshot 2024-02-22 at 13 59 12

logo should shrink?

efstajas:
also let's remove the divider between profile and settings, would look cleaner imo

Screenshot 2024-02-22 at 14 00 35

brandonhaslegs:
> logo should shrink?

Yes it should. Also it'll only shrink when you're connected to Goerli. Without that it should fit fine in there.

src/lib/components/header/header.svelte
```diff
@@ -91,7 +97,7 @@
     background-color: var(--color-background);
     transition: box-shadow 0.3s, background-color 0.5s;
     display: flex;
-    gap: 0.5rem;
+    /* gap: 0.5rem; */
```
jo-elimu, line 100:
```suggestion
```

jo-elimu (approved):

efstajas (approved):
Is there no way to have the logo shrink dynamically based on available space…? It does that on the LP already. Maybe try removing `header/drips-logo.svelte` and just using `illustrations/logo.svelte`, perhaps together with a new max-width prop. It should be possible.

Either way, also OK to merge like this, but I find shrinking the logo to an arbitrary width based on which network you're connected to a bit awkward.

jo-elimu (approved):

evvvritt:
Ah sure didn't realize we had a second logo component. Yes, shrinking now automatically:
Screenshot 2024-02-27 at 13 45 42
Screenshot 2024-02-27 at 13 46 19

jo-elimu (approved):

PR #995, opened 2024-02-22
fix: elevation color should be foreground

evvvritt:
restores proper border color, from
image
to
Screenshot 2024-02-22 at 15 14 34


brandonhaslegs:
Is this about a specific issue?

efstajas (approved):

PR #998, opened 2024-02-23
Add `windows-latest` to CI

jo-elimu:


jo-elimu:
> Any specific reason to run svelte-check on macos as well?

Just to make sure the software compiles for developers using different types of operating systems. We should also add support for developers with Windows machines (`windows-latest`), but that will require some more code changes.

jo-elimu:
@efstajas Getting "Container action is only supported on Linux": https://github.com/drips-network/app/actions/runs/8027003698/job/21930386799?pr=998



How do developers on Mac OS typically solve this?

efstajas:
Windows support should be given once #997 is merged.

I don't see how running our GH actions on mac OS in addition to Ubuntu would help ensure cross-platform compatibility of our dev stack. I could see running e.g. E2E tests on Mac OS, but they're executed within a docker container anyway. `svelte-check` is cross-platform and should return the exact same output no matter what platform it's run on — if not, that's a bug with svelte-check, not with our project.

efstajas:
I still don't understand the goal here — what's the benefit of running these actions on multiple different operating systems?

efstajas:
Closing this for now until the above ☝️ is addressed

Issue #950, opened 2024-02-12
move the "supported" line to its own line


brandonhaslegs:
![image](https://github.com/drips-network/app/assets/4406983/7cf1b4e9-0619-4d97-817d-a2d16caf306a)

brandonhaslegs:
Was this completed?

efstajas:
@brandonhaslegs Maybe the supporter pile could go into the "Total dripped" pill? like "$1234 from < supporter pill >"

PR #1007, opened 2024-02-27
feature/multiplayer: prep drip-list-card: tabbed views

evvvritt:
- split drip-list-card into a -thumblink and full version
- moved creator inside full card
- prepped tabbed view (add `?tabs` to drip list page url to view)


brandonhaslegs:
Is it possible to make this animation go in the direction of the tab? like slide/fade left when you click "Current" and slide/fade right when you click "In Voting"?

current vs in voting tab animation

evvvritt:
I considered but think it looks kinda weird/nauseating? - maybe because the direction of the list is rightward so leftward animation seems more natural.

https://github.com/drips-network/app/assets/6071219/e6681e81-c7aa-4299-9ab4-39c96174c275

brandonhaslegs:
I like that better. @efstajas ?

efstajas:
Just pushed this:

https://github.com/drips-network/app/assets/1018218/3801f877-8ec7-4f36-b738-8c931bb32e19

Wdyt?

Also fyi @evvvritt i removed the `?tabs` param, i think it's ok if we just have it always enabled for now on the feature branch. Also, we need to do something about the `Supported by` row — right now it only renders client-side, which makes the page reshuffle awkwardly on first load. should probably be inline with the total donation amount?

evvvritt:
Nice, transition looks good.
I'm agnostic on where Supported by is, @brandonhaslegs's latest design has it on its own row but I restored it next to dripped amount.

brandonhaslegs:
it should be on its own line. see here why https://github.com/drips-network/app/issues/950

evvvritt:
All devices on it's own line? And if so, should it animate-in in some way?

Same q for the mini-card we show in sections/Explore

brandonhaslegs:
I don't understand your question sorry

evvvritt:
Should "Supported by..." be on it's own line on mobile and desktop view? If so, should it animate into view because that Supporter list is fetched on page load so the layout will shift once it's loaded.

And then, worth noting that we currently use a different / "mini" drip-list-card component in drip-list-sections that appear in the Dashboard and Explore. Should that also have show "Supported by..." in it's own row.

efstajas:
I think it makes more sense for the "supported by" line to be under the "current" tab. It would further reinforce the fact that the "in voting" list is off-chain and not supportable.

brandonhaslegs:
But that seems like the supporters only apply to the current recipients. As if they'll be dropped once the voting round ends or something. Supporters apply to the whole Drip List, whether it's beinh voted on or not.

Unless I'm misunderstanding you. If you're just talking about a draft drip list, there should be no "supported by" line in a draft drip list.

Issue #983, opened 2024-02-16
allow any user to start a Continuous Donation to a Drip List


brandonhaslegs:
when you're the owner
![image](https://github.com/drips-network/app/assets/4406983/ece26117-7f37-49c4-98c4-08e2d2ba1e5e)

when you're not 
![image](https://github.com/drips-network/app/assets/4406983/b3330d2b-7b12-44a1-990e-89a1e2aba23b)

No reason for this

More context here

https://github.com/drips-network/app/assets/4406983/ba4fc9a9-9f38-46ec-8b1c-521381e26f69

This text needs to be updated when this is done.
<img width="503" alt="image" src="https://github.com/drips-network/app/assets/4406983/3ff1dd55-4a3a-46e9-a765-8492f2e5838c">

ashiskumar-1999:
Hi@brandonhaslegs, can you provide more insight on how will it work?

brandonhaslegs:
@ashiskumar-1999 I added more context above

brandonhaslegs:
> Also, the given text should replace this one Donate instantly or add them to a Drip List. Right?

No it will just be a 3rd button. Like this:
<img width="455" alt="image" src="https://github.com/drips-network/app/assets/4406983/07c9d1b7-7fb9-4802-99ed-37141dbdb6ea">

ashiskumar-1999:
Hi @brandonhaslegs, need a review on this PR. Review it once when you're available.

brandonhaslegs:
best that @evvvritt or @efstajas reviews it.

PR #1019, opened 2024-03-06
fix: added continuous donation to drip list

ashiskumar-1999:
Added the logic in the continuous donation button to allow any users to start a continuous donation to any drip list.

closes: (#983)


ashiskumar-1999:
Hi @efstajas , I'd like to continue this task. Can you give me details that we're about to change with this new feature?
Also, I have tried to query all the accounts that streaming to the given drip list but it throws an error of `Introspection is disabled on this endpoint. Enable introspection to populate your schema.` on the Apollo server playground. can you help me to find why this is happening?

ashiskumar-1999:
Hi @efstajas ,
I want to continue working on this issue. Before that, I have a question regarding the fetching of all accounts streaming to the drip list except the owner. 

- Is there any backend implementation to fetch all the accounts that stream to the drip list as `continuous donation`?because I have tried to query the accounts but didn't get any fields that take me to fetch the accounts having continuous donations.

efstajas:
Hey @ashiskumar-1999, great to hear!

Streams information is currently not available on the GQL API. Instead, you're going to have to use the `streams.store`. There is a function `fetchAccount`, which loads a particular account into the streams store, and also fetches all accounts streaming to that account. You should be able to call `fetchAccount` with a Drip List ID, and it should just automatically fetch all streams coming in to that account. After calling that function, call `getStreamsForUser` with the Drip List ID, and you should receive all relevant `outgoing` streams.

efstajas:
hey @ashiskumar-1999, sorry but someone from our team took this task on now since it's been near the top of our priorities for a while now. Hope you didn't spend too much time on it, and thanks a lot for your efforts.

PR #1046, opened 2024-03-28
More resilient FUNDING.json Comparison

holgerd77:
Hi there,
I just wanted to claim funds for the EthereumJS monorepo and couldn't get the `FUNDING.json` file to get accepted.

After some digging I realized that I had created a lower-case address which I then imported into MetaMask, and MetaMask had transformed this to a partly-capitalized address (which is ok since Ethereum addresses are case insensitive and this can be used for additional check-summing).

So I guess it is more resilient and less error-prone for `FUNDING.json` comparison to compare the lower-case versions of the file.

### DIFF
--- a/src/lib/utils/github/GitHub.ts
+++ b/src/lib/utils/github/GitHub.ts
@@ -49,7 +49,7 @@ export default class GitHub {
 
     const fundingJson = JSON.parse(fileContent);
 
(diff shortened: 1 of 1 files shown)


efstajas (approved):
Thanks a ton!

PR #1055, opened 2024-04-08
refactor list editor

efstajas:


src/lib/utils/project/GitProjectService.ts
```diff
@@ -327,97 +329,102 @@ export default class GitProjectService {
...
 
-      maintainersSplitsMetadata.push(receiver);
       receivers.push(receiver);
     }
 
+    // Adjust weights to ensure no tiny remainder
+    const MAX_WEIGHT = 1000000;
+
+    function adjustWeights<T extends { weight: number }>(input: T[]): T[] {
+      const totalWeight = Object.values(input).reduce((acc, { weight }) => acc + weight, 0);
```
jtourkos, line 408: Is `Object.values(input)` necessary since input is already an array?

src/lib/utils/project/GitProjectService.ts
```diff
@@ -327,97 +329,102 @@ export default class GitProjectService {
...
+      const remainder = MAX_WEIGHT - totalWeight;
+
+      if (remainder > 0 && remainder < 5) {
+        input[0].weight += remainder;
```
jtourkos, line 412: Wouldn't this throw if the input is empty?

src/lib/utils/project/GitProjectService.ts
```diff
@@ -327,97 +329,102 @@ export default class GitProjectService {
...
     }
 
+    // Adjust weights to ensure no tiny remainder
+    const MAX_WEIGHT = 1000000;
+
+    function adjustWeights<T extends { weight: number }>(input: T[]): T[] {
+      const totalWeight = Object.values(input).reduce((acc, { weight }) => acc + weight, 0);
+      const remainder = MAX_WEIGHT - totalWeight;
+
+      if (remainder > 0 && remainder < 5) {
```
jtourkos, line 411: Why 5? Could it be > 5?

src/lib/utils/project/GitProjectService.ts
efstajas: I think it's OK to assume it's not empty here bc a project can't have no receivers.

PR #1069, opened 2024-04-26
Fix small issues updated

evvvritt:
fixed this
- applied `text-wrap:pretty` to `<body>` (rather than `*` which overrides `whitespace:nowrap` in parent elements ¯\_(ツ)_/¯)

image


efstajas (approved):

evvvritt:
thanks for approving @efstajas, but I don't have priv's to merge without e2e passing (which are still bonked right?)

efstajas:
the tests should work again shortly, events processor upstream just has a
little bug atm. will re-trigger once it's fixed!

PR #1070, opened 2024-04-30
feat: landing page multiplayer

evvvritt:
code in the new landing page :}


brandonhaslegs:
Would be cool if the coins were animating through from left to right in this section
image

brandonhaslegs:
These should end with some kind of node (an avatar or project or something) instead of just ending lines (my bad!)
image

brandonhaslegs:
I think the typing should also "delete" the letters in the other direction before the next phrase shows up. or "select" them before they disappear to seem like they're being selected then deleted.

image

brandonhaslegs:
Feedback from Ele
- Make the graph more in the background. it's too busy up top
- create a bit more space between the animating title and the graph elements

PR #1191, opened 2024-08-23
allow CSV import when editing existing Drip List

efstajas:
Adds ability to import receivers from CSV for existing Drip Lists as well.

image

Also includes some refactoring of Drip List editing-related components and steps. Nice side effect is that the state of a Drip List edit now survives when confirming & going back to the edit step after a TX-related error.


efstajas:
CC @nya-elimu

nya-elimu:
@efstajas If we import from CSV, will the _existing_ recipients be deleted automatically? Or do we have to manually each existing recipient from the list before importing the updated CSV?

efstajas:
> @efstajas When we import from CSV, will the existing recipients be deleted automatically before the import happens?

Yes, the import will replace any existing receivers with whatever was imported. Is that what you would expect?

nya-elimu:
> Yes, as-is, the import will replace any existing receivers with whatever was imported. Is that what you would expect?

@efstajas Yes, that would be good. Because this way we won't have to manually go through the whole existing list and remove every single recipient before adding the updated list of recipients and percentages, as is the case today. So deleting all existing recipients would save us a lot of time 👍🏽

jtourkos (approved):

Issue #1180, opened 2024-08-19
Seamless wrapping & unwrapping of wrapped native tokens 


efstajas:
Currently, wrapped tokens like WETH or WFIL are just part of our default token list, and can be topped up & collected in their wrapped states. For recipients of such wrapped tokens, it'd be awesome if they were to get automatically unwrapped upon collecting.

Similarly, it'd be great if the app could automatically wrap native tokens (e.g. ETH) upon top-up or one-time donation, though this is less important because funding interactions are 1) usually done by crypto natives and 2) happen a lot less frequently than collects.

@CodeSandwich has an idea for an on-chain helper that would allow us to batch an `unwrap` interaction directly with the existing `collect` batch TX. With this in place, we could add an option to seamlessly unwrap received WETH or WFIL into their native counterparts on collect.

We should explicitly define a list of native tokens that support this auto wrap / unwrap functionality in the app, because not all wrappers share the same API. For the beginning, it should support ETH/WETH on mainnet and FIL/WFIL on Filecoin.

UX outline for collecting:
- supported wrappers appear with their non-wrapped name, but accurate ERC-20 symbol in token lists and across the app, e.g. Ethereum (WETH) or Filecoin (WFIL).
- When collecting at least one supported wrapper, a "Unwrap supported tokens" options checkbox, enabled by default, appears in the collect modal.
- If checked, the user ends up with the native counterparts of collected wrapped tokens (ETH, FIL), in their wallet, rather than the ERC-20 wrappers.

UX outline for topping up / one-time donating supported native tokens:
- BOTH the native token & wrapped counterpart appear in the token picker (because someone may have both ETH/FIL and WETH/ETH in their wallet).
- If the user chooses the native token, we automatically include the wrap interaction in the subsequent transaction.

CodeSandwich:
> We should explicitly define a list of native tokens that support this auto wrap / unwrap functionality in the app, because not all wrappers share the same API. For the beginning, it should support ETH/WETH on mainnet and FIL/WFIL on Filecoin.

AFAIR all the popular native token wrappers I saw were sharing the WETH9 API. There were some extra features, but they were all regular ERC-20s with additional `deposit()` (also triggered when doing a plain token transfer) and `withdraw(uint amount)` functions.

efstajas:
note to self: split into two tasks, one for collecting & one for topping up.

brandonhaslegs:
[...]

PR #1198, opened 2024-09-05
feat: auto unwrap on collect

jtourkos:
Resolves: https://github.com/drips-network/app/issues/1180

https://github.com/user-attachments/assets/c0728482-4c2c-4816-b5e8-3b7c242678d1

Not sure if we want this 👆 behavior regarding the token name, when toggling the auto-unwrap option.
But it might be a bit confusing having this turned off, and still show FIL and not WFIL, as I would expect as user.


src/lib/stores/wallet/network.ts
```diff
@@ -100,6 +107,11 @@ const NETWORK_TOKENS: ValueForEachSupportedChain<string> = {
   [314]: 'FIL',
 };
 
+const NETWORK_AUTO_WRAPPED_PAIRS: Partial<ValueForEachSupportedChain<AutoUnwrapPair[] | []>> = {
+  [1]: [{ name: 'Ethereum', nativeSymbol: 'ETH', wrappedSymbol: 'WETH' }],
+  [314]: [{ name: 'Filecoin', nativeSymbol: 'FIL', wrappedSymbol: 'WFIL' }],
```
jtourkos, line 112: Should these go into the `additional-tokens.json` instead?

src/lib/components/token/token.svelte
```diff
@@ -33,6 +35,22 @@
...
+  // If auto unwrap is enabled, show the native name instead of the wrapped token's name.
+  $: {
+    const autoUnwrappedTokenPair = network.autoUnwrapPairs?.find(
+      (p) => p.wrappedSymbol === token?.info.symbol,
+    );
+
+    if (autoUnwrappedTokenPair && showWrappedAsNative && tokenInfo) {
+      tokenInfo = {
+        ...tokenInfo,
+        name: autoUnwrappedTokenPair.name,
```
jtourkos, line 47: See related comment in `network.ts` about the `name` property.

brandonhaslegs:
This video you shared is expected but I agree it's going to be confusing for people. Maybe we should have a wrapped icon instead of just calling it `Bitcoin (WBTC)`

src/lib/stores/wallet/network.ts
efstajas: I think here is fine. I'd rather not add to additional-tokens bc the "token list" standard is explicitly only for ERC-20s AFAIK

efstajas (approved):

PR #1208, opened 2024-09-23
test: add utils test case

yuzuswapteam005:
add utils test case


efstajas:
Hey @yuzuswapteam005 thank you for your contribution! could you please rename the test files to end in `.unit.test.ts`? otherwise they don't get picked up by our runner. Thanks!

Issue #1181, opened 2024-08-20
RPC redundancy


efstajas:
The Drips App, Event Processor and API shouldn't go down if our main RPC provider, Infura, goes down. We should implement some RPC redundancy with a secondary RPC provider on a cheap plan with auto-scaling enabled.

The changes below are described for the app specifically, but similar logic should be put in place for Event Processor and API.

Outline:
- Sign up for a paid account with an alternative provider (e.g. Alchemy), choose a low-tier plan, but enable auto-scaling in case the fallback goes into effect during a time of elevated traffic.
- Create a generic RPC API endpoint that we direct all RPC traffic from the app to, e.g. `/api/rpc`.
- Any RPC hitting the endpoint should first go to Infura. If Infura returns an error or times out (low timeout threshold), it should retry the same request with the backup RPC.
- Return response of backup RPC or error for good if the backup is also faulty.

Important is that we should catch and forward RPC errors from main to Sentry, even if the backup saved the request, so that we get notified.

Status:
- [ ] App (https://github.com/drips-network/app/pull/1218)
- [x] Events processor
- [ ] Multiplayer API (https://github.com/drips-network/multiplayer/pull/15)
- [ ] API (https://github.com/drips-network/graphql-api/pull/28)
- [x] Setup failover account

bordumb:
Cross-posting here just so the idea isn't lost...

I saw the recent X/Twitter post that Drips was affected by Infura going down.

I've heard that some teams are using Grove as a fallback for issues like these. 

Details below. Pretty good list of others using it.

https://www.grove.city/

PR #1218, opened 2024-09-25
refactor: use the Failover Provider

jtourkos:
(partially) resolves: https://github.com/drips-network/app/issues/1181


src/lib/utils/FailoverProvider.ts
```diff
@@ -0,0 +1,95 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
efstajas, line 1: why? i don't see anywhere you couldn't just use the safer `unknown` instead of `any` in this file

jtourkos:
https://github.com/drips-network/app/pull/1263

PR #1226, opened 2024-09-30
fix(ui): rephrase "created by" to "owned by"

### LINKED ISSUE #1225: Change "Created by" to a more accurate text for Drip List owners
A Drip List's "Created by" address in the UI is not accurate because a creator can transfer the Drip List NFT to a different account after creating the Drip List. And when that happens, the text "Created by" no longer makes sense.

Rephrase to something more accurate like "Owned by" or "Controlled by"/"Managed by"

jo-elimu:
Resolves #1225

<kbd></kbd>

### DIFF
--- a/src/lib/components/drip-list-card/drip-list-card.svelte
+++ b/src/lib/components/drip-list-card/drip-list-card.svelte
@@ -244,7 +244,7 @@
       {/if}
       {#if !listingMode}
         <div class="flex gap-2">
-          Created by <IdentityBadge
+          Owned by <IdentityBadge
             showAvatar={true}
             showIdentity={true}
             address={listOwner?.address ?? votingRound?.publisherAddress ?? unreachable()}


jo-elimu:
@efstajas

jo-elimu:
Getting a TimeoutError for the E2E tests:
```
app-1  |  FAIL  src/e2e-tests/top-up-create-stream.e2e.test.ts > app > streams and balances > search > opens the searchbar
app-1  | TimeoutError: locator.click: Timeout 30000ms exceeded.
app-1  | Call log:
app-1  |   - waiting for locator('data-testid=search-button')
```

efstajas:
Thanks @jo-elimu, change LGTM 👍 The E2E tests are currently broken, that's expected.

efstajas (approved):

efstajas:
Actually, now thinking that maybe "Owned by" in context of the total value being shown might imply that the value is somehow owned by the address shown, especially given lists being NFTs is not immediately clear just by looking at one. Maybe "Controlled by" is better, though it sounds a bit technical. @brandonhaslegs you have an opinion on this?

brandonhaslegs:
Can it remain "created by" until it's actually transferred?

jo-elimu:
> Actually, now thinking that maybe "Owned by" in context of the total value being shown might imply that the value is somehow owned by the address shown, especially given lists being NFTs is not immediately clear just by looking at one. Maybe "Controlled by" is better, though it might sounds a bit technical.

@efstajas Yes, I think you are right that "Owned by" can confuse people to think that the Drip List owner owns the funds in the Drip List. So something like "Controlled by," "Operated by," or "Managed by," might be better. I think that "Managed by" is probably the least technical term and the best option, so I adjusted the PR here: https://github.com/drips-network/app/pull/1226/commits/ead970e48a5faafb12515c8786646e27664f5a45

jo-elimu:
> to make it dynamic, you could just add `previousOwnerAddress` to the Drip List Badge fragment, and display "Created by" if that value is `0x0000000000000000000000000000000000000000`, or "Controlled by" if it's another address. Would you be up for making that chance?

@efstajas Will that solution cover the edge-case where a Drip List changes owner 3 times or more? If the Drip List ownership is transferred to a 2nd account, it would be correct. But if the 2nd owner transfers ownership to a 3rd account, would it still be correct to assume that the `previousOwnerAddress` is the creator?

efstajas:
@jo-elimu No, if it were transferred a third time, `previousOwnerAddress` would be the address of the 2nd owner. But it shouldn't matter for this if all we're trying to check is whether the current owner *created* the list or not. Unless the previous owner is the null address, it hasn't been created by the current owner (unless the list was transferred back to the original creator, but that's an edge case I think we can ignore).

PR #1262, opened 2024-10-14
explicit errors if gasless owner update fails

efstajas:
Currently, if the gasless owner update in the project claim flow fails to kick off (e.g. Gelato Relay down) or the task fails to conclude (e.g. we ran out of balance to sponsor gas with), the user is just left sitting on the "Waiting for verification to finalize..." screen while the app polls the API in anticipation of the owner updating. Which of course will never happen in those cases. This adds explicit checks for a successful gasless owner update call, and checks the task state on Gelato for failures. In both cases, the user will see "friendly" errors asking for them to either try again or reach out to us on Discord.

### DIFF
--- a/src/lib/flows/claim-project-flow/steps/poll-api/poll-api.svelte
+++ b/src/lib/flows/claim-project-flow/steps/poll-api/poll-api.svelte
@@ -13,6 +13,7 @@
   } from './__generated__/gql.generated';
   import filterCurrentChainData from '$lib/utils/filter-current-chain-data';
   import network from '$lib/stores/wallet/network';
+  import assert from '$lib/utils/assert';
 
   const dispatch = createEventDispatcher<StepComponentEvents>();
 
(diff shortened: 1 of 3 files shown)


jtourkos (approved):

src/lib/flows/claim-project-flow/steps/poll-api/poll-api.svelte
```diff
@@ -46,6 +47,38 @@
       }
     `;
 
+    assert($context.gaslessOwnerUpdateTaskId, 'Gasless owner update task ID is missing');
```
mhgbrown, line 50: Does assert eventually produce some sort of global error display for the user?

src/lib/flows/claim-project-flow/steps/poll-api/poll-api.svelte
```diff
@@ -46,6 +47,38 @@
...
+
+    // First, wait for Gelato Relay to resolve the update task.
+    await expect(
```
mhgbrown, line 53: Love this function!

mhgbrown (approved):
Just asked a little question about assert

src/lib/flows/claim-project-flow/steps/poll-api/poll-api.svelte
efstajas: Yep, any error thrown within the context of a `Stepper` will be caught by the stepper component and trigger an `Error` step that just stringifies the error and shows it to the user.

PR #1263, opened 2024-10-14
refactor: integrate FailoverProvider

jtourkos:


src/routes/api/rpc/[provider]/[network]/+server.ts
```diff
@@ -0,0 +1,83 @@
...
+  const headers = new Headers();
+
+  if (provider === 'infura') {
+    const infuraKey = getOptionalEnvVar('INFURA_KEY');
+    assert(infuraKey, 'INFURA_KEY is required');
+    rpcUrl = `https://${network}.infura.io/v3/${infuraKey}`;
+  } else if (provider === 'alchemy') {
+    const alchemyKey = getOptionalEnvVar('ALCHEMY_KEY');
+    assert(alchemyKey, 'ALCHEMY_KEY is required');
+    rpcUrl = `https://${alchemyNetworkMap[network as keyof typeof alchemyNetworkMap]}.g.alchemy.com/v2/${alchemyKey}`;
```
Comment (efstajas) [inline src/routes/api/rpc/[provider]/[network]/+server.ts:50]: The typecast here could be unsafe if `provider` is `alchemy` but `network` is `filecoin-mainnet`. Generally I think typecasting should be avoided unless absolutely necessary. I would assert explicitly here that `network` is included in `alchemyNetworkMap`

.env.template
```diff
@@ -1,8 +1,12 @@
 # URL to use for fetcing IPFS content. Set to https://drips.mypinata.cloud to use public Drips gateway.
 PUBLIC_PINATA_GATEWAY_URL=string
 
-# Infura key for connecting to Ethereum networks. You can create a free Infura account for development.
+# Optional Infura key.
 INFURA_KEY=string
+# Optional Alchemy key.
+ALCHEMY_KEY=string
+# Optional Filecoin key.
+FILECOIN_KEY=string
```
efstajas, line 9: could you just add a sentence explaining that values for at least one of the providers are required?

efstajas (changes requested):

src/routes/api/rpc/[provider]/[network]/+server.ts
Comment (efstajas) [reply src/routes/api/rpc/[provider]/[network]/+server.ts:50]: oh also, i think the entire block would be neater as a `switch` statement

efstajas (approved):

PR #1266, opened 2024-10-16
Fix be smart about column names in csv import

mhgbrown:
This PR changes the CSV parsing logic so that the column order does not matter for files with headers. The header names must now be specified explicitly in the code. It also changes the validation logic for ENS names so that invalid ones will appear in the list error messages and not just "Something terrible happened". The PR also adds parsing tests and associated CSV files. Executed `npm run format`


src/lib/flows/import-from-csv/parse-upload.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Determine if at least on of the specified header names are in the header
+ * of the given parsed CSV file.
+ *
+ * @param parsedFile A parsed CSV file.
+ * @param csvHeaders A list of header names.
+ * @returns True if at least on the specified header names exists in the
+ *  header, false otherwise.
+ */
+function hasHeader(parsedFile: Array<Array<string>>, csvHeaders: Array<string>): boolean {
```
efstajas, line 12: not a hard rule but we usually use e.g. `string[]` syntax to denote array types, rather than `Array<string>`. Bit more compact 👍

efstajas (approved):
nice!

src/lib/flows/import-from-csv/parse-upload.ts
mhgbrown: Ok, I'll update this

PR #1278, opened 2024-10-17
fix: #1276

### LINKED ISSUE #1276: Blog should not be accessible on Filecoin deployment
We need to start redirecting `/blog/` routes to the main deployment if in "alternative chain mode".

efstajas:
While we have one deployment per chain, we don't want to create duplicates of the LP, privacy pages, and blog (incl posts). They should be exclusively served by the main deployment. This was already implemented, but not yet for blog routes.

- Move `alternativeChainMode` config from env to static network config
- Make it so that when someone tries to access blog posts on a deployment in `alternativeChainMode`, the request gets rewritten to the main (`https://drips.network`) deployment

Resolves #1276


mhgbrown (approved):
Looking good!

PR #1290, opened 2024-10-21
1238 add custom share modal

mhgbrown:
This PR expands the functionality of the share button to open a share modal when the current device is not a touchscreen.
Screenshot 2024-10-21 at 13 29 36


brandonhaslegs:
Since Signal got booted, the balance is off and it looks dorky. Can add an "email" button option

Here's the email icon (unless there is already one in the design system?)

```<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
<path fill-rule="evenodd" clip-rule="evenodd" d="M5 4.75C3.61929 4.75 2.5 5.86929 2.5 7.25V16.75C2.5 18.1307 3.61929 19.25 5 19.25H19C20.3807 19.25 21.5 18.1307 21.5 16.75V7.25C21.5 5.86929 20.3807 4.75 19 4.75H5ZM3.5 7.25C3.5 6.42157 4.17157 5.75 5 5.75H19C19.8284 5.75 20.5 6.42157 20.5 7.25V8.12934L12.8266 13.197C12.3252 13.5281 11.6747 13.5281 11.1733 13.197L3.5 8.12939V7.25ZM3.5 9.32779V16.75C3.5 17.5784 4.17157 18.25 5 18.25H19C19.8284 18.25 20.5 17.5784 20.5 16.75V9.32774L13.3777 14.0315C12.5421 14.5833 11.4578 14.5833 10.6222 14.0315L3.5 9.32779Z" fill="black"/>
</svg>```

Also I'd say keep the subject empty and just put the link in the body.

File: src/lib/flows/share/steps/share-url.svelte [hunk 1]
```diff
@@ -0,0 +1,213 @@
...
+  import CheckCircleIcon from '$lib/components/icons/CheckCircle.svelte';
+  import CopyIcon from '$lib/components/icons/Copy.svelte';
+  import EmailIcon from '$lib/components/icons/Email.svelte';
+  import { fade } from 'svelte/transition';
+  import { browser } from '$app/environment';
+  import { type ShareOption } from '../share-steps';
+
+  export let url: string = '';
+  export let downloadableImageUrl: string = '';
+  export let text: string = '';
```
efstajas, line 22: Small nitpick, but I think it'd be nicer to have these optional string props defined like this instead:

`export let prop: string | undefined = undefined` 

It communicates clearer that the prop is actually optional, and also makes it so that you don't have to explicitly pass an empty string when you actually want to pass nothing.

src/lib/flows/share/steps/share-url.svelte
```diff
@@ -0,0 +1,213 @@
...
+                </span>
+              {:else}
+                <span transition:fade={{ duration: 200 }}
+                  ><LinkIcon style="fill: currentColor" /></span
+                >
+              {/if}
+            </div>
+            Copy link
+          </div>
+        </Button>
```
efstajas, line 131: i think would be slightly nicer if the copy link button was a separate `copy-link-button.svelte` component which is only imported here, just so that the `hovering` state which is only for the copy linking button is more neatly encapsulated and clearly only for the copy link button

src/lib/flows/share/steps/share-url.svelte
```diff
@@ -0,0 +1,213 @@
...
+  .share-options__options :global(button),
+  .share-options__options :global(a) {
+    width: calc(50% - 1rem);
+  }
+
+  .share-options__options :global(button .inner),
+  .share-options__options :global(a .inner) {
+    /* yoiks */
+    justify-content: start !important;
+  }
```
efstajas, line 191: why are the `:global` and `!important` needed here?

src/lib/components/share-button/share-button.svelte
```diff
@@ -1,75 +1,70 @@
...
+    try {
+      const img = new Image();
+      img.src = url;
+      await img.decode();
+    } catch (error) {
+      // we don't really care to make a big fuss about this
+      // eslint-disable-next-line no-console
+      console.error('Error preloading image', error);
+    }
+  }
```
efstajas, line 31: nice 👍

efstajas (changes requested):
looks awesome ✨ few very minor nitpicks

src/lib/flows/share/steps/share-url.svelte
mhgbrown: Makes, total sense will do!

src/lib/flows/share/steps/share-url.svelte
mhgbrown: Done!

efstajas: If URL is required, just remove the default — then it'll actually be treated as a required prop by the compiler

mhgbrown: Ah hah! Will update

src/lib/components/copy-link-button/copy-link-button.svelte
```diff
@@ -0,0 +1,67 @@
+<script lang="ts">
+  import Button from '$lib/components/button/button.svelte';
+  import LinkIcon from '$lib/components/icons/Link.svelte';
+  import CheckCircleIcon from '$lib/components/icons/CheckCircle.svelte';
+  import CopyIcon from '$lib/components/icons/Copy.svelte';
+  import { fade } from 'svelte/transition';
+
+  export let url: string = '';
```
efstajas, line 8:
```suggestion
  export let url: string;
```

efstajas (approved):

PR #1293, opened 2024-10-22
Fake progress bar while waiting for owner update, reduce unnecessary owner update TXs

### LINKED ISSUE #1249: Fake progress bar on transaction screen
FVM tends to be a bit slow, so we want to display some progress on the Transaction screen during the project claim flow. There should be a "fake" progress bar that assumes a reasonable duration and resolves fully once the underlying task is complete.

There should be a progress bar displayed in the project claim flow while:
- We're waiting for the project owner update from the Oracle
- We're waiting for the final claim TX being confirmed on-chain

### LINKED ISSUE #1319: Don't kick off another Gelato Relay TX if owner is already the right one
Currently, we kick off a `requestUpdateOwner` TX every time the funding.json file is successfully validated client-side during the project claim flow. We should check the current on-chain owner via API first, and only kick this TX off if it doesn't already match the owner we want to set.

If no TX is necessary, the "Waiting for verification to finalize" step should skip waiting for the Gelato task to resolve.

efstajas:
Adds a new progress bar component, and a new feature to the `transact-step` that allows adding "external" transactions that resolve with some arbitrary promise. Adds a fake progress bar for when we're waiting for Gelato Relay to do its thing in the project claim flow.

Also, since this logic is tightly related, makes it so that we don't kick off unnecessary gasless TXs if the owner of the repo is already the right one at the beginning of the flow (for example, if the user previously dropped out of the flow halfway through).

Resolves #1249
Resolves #1319


src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -38,23 +42,30 @@
...
-      | 'failed'
-      | 'cancelled'
-      | 'rejected'
-      | 'retrying'
-      | 'finalizing'
-      | null;
-  }
+  type TransactionTimelineItem =
+    | {
+        external: false;
```
jtourkos, line 47: Generic minor suggestion / personal preference for booleans: rename to `isExternal`.

jtourkos (approved):

PR #1305, opened 2024-10-25
fix ens

### LINKED ISSUE #1302: Profile and ENS names not loading
Looks like the ENS metadata is there but the profile pic and name are gone in the whole app for every profile.

Image
Image

efstajas:
Resolves #1302

Not 100% quite sure what caused the problem in the end. I switched to using an explicit `EnsResolver` for resolving Avatars, as well as explicitly defining ENS addresses for networks. That seems to have done the trick. Also updates Ethers minor version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10055,9 +10055,9 @@
       }
     },
     "node_modules/ethers": {
-      "version": "6.13.3",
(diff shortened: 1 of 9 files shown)


mhgbrown (commented):
Looks good to me. It looks like there are some type check errors?

PR #1331, opened 2024-10-30
Improve gasless owner update endpoint validations

efstajas:
The endpoint is intended only for requesting owner updates of unclaimed repos as part of the project claim flow.

This adds some additional validations which would make it a lot harder to spam the endpoint with nonsense requests:
- Check that the project is actually unclaimed
- Skips Gelato requests and just returns 200 if the owner on-chain already matches what's in the funding.json
- Sets a key on Redis that includes the task ID for a given update task of the same repo. If that key is present, checks the state of the task. If the task is still in flight, returns the same task ID as the in-flight request. This prevents creating multiple tasks that update the same repo owner at the same time.


jtourkos (approved):

efstajas:
Included with #1368

PR #1367, opened 2024-11-29
1295 add a warning not to send funds to exchange managed addresses

mhgbrown:
This PR adds a warning for eth address entry situations telling the user to add only self-custodial addresses.
Screenshot 2024-11-29 at 19 31 21

### DIFF
--- a/src/lib/components/annotation-box/custodial-warning.svelte
+++ b/src/lib/components/annotation-box/custodial-warning.svelte
@@ -8,15 +8,11 @@
 
   export let dismissableId: string = 'custodial';
 
-  let dismissed = browser ? dismissablesStore.isDismissed(dismissableId) : false;
+  $: dismissed = browser ? $dismissablesStore.includes(dismissableId) : false;
 
   async function handleDismiss() {
     dismissablesStore.dismiss(dismissableId);
   }
-
-  dismissablesStore.subscribe(() => {
-    dismissed = dismissablesStore.isDismissed(dismissableId);
-  });
 </script>
 
 {#if !dismissed}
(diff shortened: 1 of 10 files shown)


efstajas (changes requested):
could you check what's up with the period? 

image

also, i think `slide` as a transition might look neater for dismissing!

efstajas:
note to self: forgot to push mobile layout for the new gasless badge

src/lib/components/annotation-box/custodial-warning.svelte
```diff
@@ -0,0 +1,45 @@
...
+
+  let dismissed = browser ? dismissablesStore.isDismissed(dismissableId) : false;
+
+  async function handleDismiss() {
+    dismissablesStore.dismiss(dismissableId);
+  }
+
+  dismissablesStore.subscribe(() => {
+    dismissed = dismissablesStore.isDismissed(dismissableId);
+  });
```
efstajas, line 19: no need for manually subbing to `dismissablesStore`! Should be possible to just do `{#if $dismissablesStore.includes(dismissableId)` on line 22

src/lib/components/annotation-box/custodial-warning.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+  import AnnotationBox from './annotation-box.svelte';
+  import ThumbsUpIcon from '$lib/components/icons/ThumbsUp.svelte';
+  import Button from '../button/button.svelte';
+  import dismissablesStore from '$lib/stores/dismissables/dismissables.store';
+  import { browser } from '$app/environment';
+  import { slide } from 'svelte/transition';
+
+  export let dismissableId: string = 'custodial';
```
efstajas, line 9: best to remove the default value IMO

efstajas (changes requested):

efstajas (approved):

efstajas (approved):

PR #1368, opened 2024-11-29
Gasless claiming & collecting

efstajas:
Adds the ability to enable entirely gasless claiming & collecting via Gelato Relay on specific networks for EOAs (disabled in Safe mode) via standard EIP-712 signatures against Caller's `callSigned` fn.

The project claim flow is already halfway gasless — the 1st transaction has been relayed via Gelato forever, and doesn't require a user sig because `requestOwnerUpdate` is a public fn. This also makes the 2nd transaction, which does require a user signature, gasless.

Availability of gasless claiming is set via the `gaslessClaimAndConnect` key within the network config. It's currently enabled for Sepolia and Filecoin Mainnet.

There's also a new user-facing setting in the "advanced" category for turning gasless txs on / off on supported networks.

High-level changes:
- A new `gasless.store.ts` that tracks the users preference from settings screen. True by default, unless on unsupported network or connected to Safe.
- A new `gasless` property for `transact` payload transactions. If true, transact step will request an EIP-712 signature of the transaction targeting `callSigned` on Caller instead of proposing the transaction directly to the wallet. Set to the value of `gasless.store.ts` for project claiming & collect flows.
- A new `/api/gasless/call` endpoint. It thoroughly validates that the transaction is for one of the two currently supported gasless interactions (claiming a project or collecting earnings) by decoding the calldata and even ensuring the project in question is currently unclaimed.

Also includes changes that greatly tighten down the `gasless-owner-update` endpoint, which will now prevent multiple requests in-flight for the same repo with a block key on Redis, as well as verify that the project in question is currently unclaimed.


src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -318,6 +325,121 @@
...
+          };
+
+          const sig = await signer.signTypedData(domain, types, payload);
+
+          updateTransactionTimelineStatus(executingTx, {
+            status: 'pending',
+            message: 'Submitting transaction',
+          });
+
+          const gaslessCallRes = await fetch('/api/gasless/call', {
```
mhgbrown, line 390: Now that I have `fetch` on the brain and I see that you check the res below, should you check for ok here as well?

src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -318,6 +325,121 @@
...
+            },
+            600000,
+            2000,
+          );
+
+          if (gaslessCallExpectation.failed) {
+            throw new Error('The gasless call did not resolve within the expected timeframe.');
+          }
+
+          break;
```
mhgbrown, line 440: What's this breaking out of?

src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -318,6 +325,121 @@
...
+        }
+
```
mhgbrown, line 442: And just noticing the `<script>` here seems long. Is there an opportunity to break it up into bite-sized chunks?

src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -671,6 +797,20 @@
...
+                <!-- Gasless indicator -->
+                {#if transactionStatusItem.gasless}
+                  <div class="gasless-notice">
+                    <Gas />
+                    Gas fee $0.00
+                    <div class="gasless-badge">
+                      PAID FOR BY
+                      <div class="logo-wrapper">
+                        <Logo style="fill: var(--color-primary-level-6)" />
+                      </div>
```
mhgbrown, line 809: 💙

src/routes/api/gasless/call/+server.ts
```diff
@@ -0,0 +1,358 @@
...
+        break;
+      }
+      default: {
+        if (!(repoDriverParseRes.name === 'split' || repoDriverParseRes.name === 'collect'))
+          return false;
+        if (repoDriverParseRes.args[0].toString() !== projectAccountId) return false;
+      }
+    }
+
+    break;
```
mhgbrown, line 186: This looks like it's going to prevent iteration past index 0. Is that desired?

src/routes/api/gasless/call/+server.ts
```diff
@@ -0,0 +1,358 @@
...
+  )?.[1];
+  assert(contractAbi, 'Invalid contract name');
+
+  const iface = new Interface(contractAbi);
+
+  const parseRes = iface.parseTransaction({ data: calldata });
+  assert(parseRes, 'Invalid calldata');
+
+  switch (contractName) {
+    case 'CALLER': {
```
mhgbrown, line 286: Did you mean to add more cases here? This is just an `if` otherwise.

src/routes/api/gasless/call/repo-owner-update/+server.ts
```diff
@@ -44,7 +83,55 @@ export const POST: RequestHandler = async ({ request }) => {
...
+
+  if (chainData.owner.address.toLowerCase() === currentFundingJsonAddress?.toLowerCase()) {
+    return new Response('{ "taskId": undefined }');
+  }
+
+  const blockKey = `${network.name}-ownerUpdateRequest-${forge}-${projectName}`;
+  const blockRecordTaskId = await redis.get(blockKey);
+
+  if (blockRecordTaskId) {
+    const taskStatusRes = await fetch(`/api/gasless/track/${blockRecordTaskId}`);
```
mhgbrown, line 121: Calling out a `fetch` without an ok check.

mhgbrown (commented):
FREE GAS! Super cool! I left some comments and questions here and there

src/lib/components/stepper/components/transact-step.svelte
efstajas: yes, prob best to check explicitly 👍 will add that

src/lib/components/stepper/components/transact-step.svelte
efstajas: #1369

src/routes/api/gasless/call/repo-owner-update/+server.ts
efstajas: added check 👍

src/routes/api/gasless/call/+server.ts
efstajas: actually, the entire logic is built with having multiple possible target contracts in the future (`SUPPORTED_CONTRACT_ABIS`, taking in `contractName` etc.), and while there's currently only Caller, I think that's fine. I think the likelihood that we want to make many more things gasless in the near future on cheap networks is pretty high. editing splits, updating metadata etc. would all go against either NFTDriver or RepoDriver, not caller, because they're not batches, so we'd need to add more contracts here.

src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -48,6 +53,7 @@
         title: string;
         message: string;
         txUrl?: string;
+        gasless?: boolean;
```
jtourkos, line 56: minor suggestion/personal preference: rename to `isGasless`

src/lib/components/stepper/components/transact-step.svelte
```diff
@@ -318,6 +325,123 @@
...
+            method: 'POST',
+            body: JSON.stringify({
+              targetContractAddress: to,
+              callData: data,
+              payload,
+              eip712Signature: sig,
+            }),
+          });
+          if (!gaslessCallRes.ok)
+            throw new Error(`Failed to submit gasless call: ${await gaslessCallRes.text()}`);
```
jtourkos, line 400:
```suggestion
});

          if (!gaslessCallRes.ok) {
            throw new Error(`Failed to submit gasless call: ${await gaslessCallRes.text()}`);
          }
          
```

jtourkos (approved):

PR #1376, opened 2024-12-02
1206 error responses occasionally being cached by cached utility (Revert Reversion)

mhgbrown:
This PR attempts to limit the incidence of caching non-OK responses via the `cached` utility and `fetch`.  Getting blog posts is method-ized and reused in places where code was being duplicated. I wanted to do the same with the `api/tlv` endpoint (and re-use `getCmcPrices`!), but didn't feel comfortable going all in without a working Etherscan api connection.

I also ran `npm run format`, which re-formatted [src/lib/utils/puppeteer.ts](https://github.com/drips-network/app/pull/1363/files#diff-8a85ae2a6d8d4a5cf81c8b8450944d2d215d5392d6771ed8fc536a3ad453817c)...I was just working on that so it's surprising to me that it needed re-formatting ¯\\_(ツ)_/¯

Reverts and builds upon drips-network/app#1372


src/routes/api/tlv/+server.ts
```diff
@@ -27,12 +27,24 @@ export const GET = async ({ fetch }) => {
...
-  ).json();
-  const dripsTokenHoldings = etherscanTokensResponseSchema.parse(dripsTokenHoldingsRes.result);
+  const driptsTokenHoldingRes = await fetch(
+    `https://api.etherscan.io/api?module=account&action=addresstokenbalance&address=0xd0Dd053392db676D57317CD4fe96Fc2cCf42D0b4&page=1&offset=100&apikey=${etherscanApiKey}`,
+  );
+  if (!driptsTokenHoldingRes.ok) {
+    const errorContent = await driptsTokenHoldingRes.text();
+    // eslint-disable-next-line no-console
+    console.error('Etherscan returned error response', errorContent);
+    return new Response('[]', { headers: { 'Content-Type': 'application/json' } });
```
efstajas, line 37: Why return an empty array? it usually returns a number as a string. on the actual `default-explore-page.svelte` it checks for `typeof tlv === 'number'`. I think ideally:

- this endpoint returns either nothing or an error 500 with empty body in case there's an etherscan error
- the logic calling in the TLV value catches an error response from this endpoint, and returns `null` in that case
- `default-explore-page.svelte` explicitly accepts either `number` or `null` as `tlv` prop, and conditionally renders the TLV `if tlv !== null`

wdyt?

efstajas (changes requested):

src/routes/api/tlv/+server.ts
mhgbrown: Yes down to change! I was following the logic in the previous lines that return an empty array if there's no etherscan key. It looks like it eventually ends up in a Intl formatting method which returns $0.0 when given an empty array...weird! 

I think returning null is great in these cases. If the tlv endpoint returns a 500, the whole loading blows up as it is currently.

efstajas: I see, didn't catch that! Yeah, if there's no etherscanApiKey i think it also should be returning null and handle it the same way as when there's an error — produce a server-side log, and gracefully hide the TLV card.

mhgbrown: Yeah that makes total sense, will do

efstajas (approved):

PR #1378, opened 2024-12-03
Change filecoin settlement to weekly on thursday

mhgbrown:
This PR updates wording to accurately reflect Filecoin settlement.

### DIFF
--- a/src/lib/flows/claim-project-flow/steps/review/review.svelte
+++ b/src/lib/flows/claim-project-flow/steps/review/review.svelte
@@ -40,6 +40,7 @@
   import filterCurrentChainData from '$lib/utils/filter-current-chain-data';
   import network from '$lib/stores/wallet/network';
   import { mapSplitsFromListEditorData } from '$lib/components/splits/utils';
+  import sanitize from 'sanitize-html';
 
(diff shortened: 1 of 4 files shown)


src/lib/flows/claim-project-flow/steps/review/review.svelte
```diff
@@ -250,9 +250,7 @@
           <span class="typo-text-bold">Drips dashboard</span>.</UlIconLi
         >
         <UlIconLi icon={TokenStreamsIcon}
-          >Future incoming funds will be split to your recipients <span class="typo-text-bold"
-            >monthly</span
-          >, on the last Thursday of every month.</UlIconLi
+          >{@html network.settlement.recipientsExplainerHtml}</UlIconLi
```
efstajas, line 253: We got burned once by `@html` resulting in what some might call a JS injection vulnerability 😇 since then, always feels better to wrap anything we pass to `@html` in `sanitize`. There's already a function for that available. I know in this case nothing user-supplied is being rendered, but better safe than sorry.

efstajas (changes requested):

efstajas (approved):

PR #1389, opened 2024-12-11
1102 populate maintainer dependency splits for projects from csv

mhgbrown:
This PR brings changes to allow users to edit their maintainers and dependencies by uploading a CSV file. It also brings the same to the project claim process. It fixes an issue where a cached version of the FUNDING.json would be returned during verification instead of the updated on.

**Flows Updated**
Claim Project
* Configure dependencies
* Configure maintainers

Edit project splits
* Edit dependency list
* Edit mantainters list


jtourkos (approved):
LGTM 👍

PR #1406, opened 2024-12-21
Generate a random Drip List ID when creating a new Drip List

mhgbrown:
Removes the count of Drip Lists when determining a salt for a new Drip List.

### DIFF
--- a/src/lib/utils/calc-salt.ts
+++ b/src/lib/utils/calc-salt.ts
@@ -1,0 +1,10 @@
+import { ethers } from 'ethers';
+
+export default function calculateSaltFromAddress(seedConstant: string, address: string) {
+  const hash = ethers.keccak256(
+    ethers.AbiCoder.defaultAbiCoder().encode(['string'], [seedConstant + address]),
+  );
+  const randomBigInt = ethers.toBigInt('0x' + hash.slice(26));
+
+  return BigInt(randomBigInt.toString()) & BigInt('0xFFFFFFFFFFFFFFFF');
+}
--- a/src/lib/utils/calc-salt.ts
+++ b/src/lib/utils/calc-salt.ts
@@ -4,7 +4,10 @@ export default function calculateSaltFromAddress(seedConstant: string, address: 
   const hash = ethers.keccak256(
     ethers.AbiCoder.defaultAbiCoder().encode(['string'], [seedConstant + address]),
   );
-  const randomBigInt = ethers.toBigInt('0x' + hash.slice(26));
+  const addressBigInt = ethers.toBigInt('0x' + hash.slice(26));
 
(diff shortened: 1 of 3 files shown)


src/lib/utils/driplist/DripListService.ts
```diff
@@ -368,25 +346,13 @@ export default class DripListService {
     return ipfsHash;
   }
 
-  // We use the count of *all* NFT sub-accounts to generate the salt for the Drip List ID.
-  // This is because we want to avoid making HTTP requests to the subgraph for each NFT sub-account to check if it's a Drip List.
-  private _calcSaltFromAddress = (address: string, listCount: number): bigint /* 64bit */ => {
+  // Create random salt from address
```
jtourkos, line 349: Are we sure this generates a _random_ number?
I think for the same address it will produce the same output, no?

jtourkos (changes requested):

src/lib/utils/driplist/DripListService.ts
mhgbrown: We're not sure and it looks like it doesn't based on the error I'm receiving. Will look into it!

src/lib/utils/calc-salt.ts
```diff
@@ -0,0 +1,13 @@
+import { ethers } from 'ethers';
+
+export default function calculateSaltFromAddress(seedConstant: string, address: string) {
+  const hash = ethers.keccak256(
+    ethers.AbiCoder.defaultAbiCoder().encode(['string'], [seedConstant + address]),
+  );
+  const addressBigInt = ethers.toBigInt('0x' + hash.slice(26));
+
+  const randomBytes = ethers.randomBytes(32);
```
jtourkos, line 9: This definitely introduces randomness now 👍
But, why not simplifying to something simple like:

```ts
export default function generateRandomNumber() {
  const randomBytes = ethers.randomBytes(32);
  return BigInt(ethers.hexlify(randomBytes)) & BigInt('0xFFFFFFFFFFFFFFFF');
}
```

mhgbrown: No reason at all. Just not trying to shake the tree too much because I don't know what constraints the seed should have. Will update!

src/lib/utils/driplist/DripListService.ts
```diff
@@ -108,7 +108,7 @@ export default class DripListService {
       items,
     );
 
-    const salt = this._calcSaltFromAddress(this._ownerAddress);
+    const salt = this._calcSaltFromAddress();
```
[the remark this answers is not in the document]
mhgbrown: No real important reason - just trying to keep the integrity of the class's internal API. I'll just use the utility directly

PR #1415, opened 2025-01-09
feat: search

efstajas:
Adds integration with `meilisearch` for searching Projects and Drip Lists. Also adds a new shortcut for brining up search: `cmd-k`

The search bar:
- makes a request to `meilisearch` on`Projects` and `Drip lists` indexes
- also looks up ens if the search input includes a `.`
- also inserts a link to a project if the search input is a URL to a valid & existing github repo


jtourkos (approved):

Dockerfile
```diff
@@ -43,6 +43,9 @@ ARG PUBLIC_BASE_URL
 ARG MULTIPLAYER_API_URL
 ARG MULTIPLAYER_API_ACCESS_TOKEN
 
+ARG MEILISEARCH_HOST
+ARG MEILISEARCH_API_KEY
+
```
mhgbrown, line 48: Can this also appear in the `.dev` version of this file and point to somewhere remote? If so I think we should add it! Can it also be added to the `.env.template`?

src/lib/components/search-bar/search.ts
```diff
@@ -1,149 +1,112 @@
...
-      type: SearchItemType.REPO;
-      matchStrings: MatchStrings;
-      item: {
-        forge: string;
-        repoName: string;
-        username: string;
-        url: string;
-      };
-    };
+  const gitHubUrl = isValidGitUrl(q) ? q : `https://github.com/${q}`;
```
mhgbrown, line 22: `isValidGitUrl` is probably not expensive, but you might save the return value from above for the check here anyways

src/lib/components/search-bar/search.ts
```diff
@@ -1,149 +1,112 @@
...
+  );
 
-  return {
-    primary: name,
-    secondary: symbol,
-    tertiary: address,
-  };
-}
+  if (!gitUrlInResults) {
+    const repoRes = await fetch(`/api/github/${encodeURIComponent(gitHubUrl)}`);
```
mhgbrown, line 29: Check for `repoRes.ok` too?

src/routes/api/multiplayer/[...path]/+server.ts
```diff
@@ -7,14 +8,17 @@ export const fallback: RequestHandler = async ({ request, params, url, fetch })
 
   const searchParams = url.searchParams.toString();
 
-  const response = await fetch(`${MULTIPLAYER_API_URL}/${params.path}?${searchParams}`, {
-    method: request.method,
-    body: body || undefined,
-    headers: {
-      Authorization: `Bearer ${MULTIPLAYER_API_ACCESS_TOKEN}`,
-      'Content-Type': 'application/json',
+  const response = await fetch(
```
Comment (mhgbrown) [inline src/routes/api/multiplayer/[...path]/+server.ts:11]: Check `response.ok` here too?

src/routes/api/search/[...path]/+server.ts
```diff
@@ -0,0 +1,36 @@
...
+if (!dev && !MEILISEARCH_HOST) {
+  throw new Error('MEILISEARCH_HOST env var is required in production!');
+}
+
+export const POST = async ({ request, url }) => {
+  const path = url.pathname.replace('/api/search/', '');
+
+  if (!MEILISEARCH_HOST) return error(500, 'MEILISEARCH_HOST env var is required for search');
+
+  const result = await fetch(`${stripTrailingSlash(MEILISEARCH_HOST)}/${path}`, {
```
Comment (mhgbrown) [inline src/routes/api/search/[...path]/+server.ts:18]: You know what I'm going to write...lol
Check `result.ok`?

Dockerfile
efstajas: forgot about both of these, good catch! will add

src/lib/components/search-bar/search.ts
efstajas: i guess i should just wrap this whole thing in a `catch` so that it "gracefully" doesn't return the repo result if there's any error with anything, but doesn't completely break search. will do that

src/routes/api/search/[...path]/+server.ts
Comment (efstajas) [reply src/routes/api/search/[...path]/+server.ts:18]: don't think it's necessary, it just proxies the request and also forwards `.status` property. Up to the client-side code to interpret that (which it does by failing the parsing and displaying an error in the search results)

src/lib/components/search-bar/search.ts
efstajas: ah that second check is completely unnecessary actually

src/routes/api/multiplayer/[...path]/+server.ts
Comment (efstajas) [reply src/routes/api/multiplayer/[...path]/+server.ts:11]: honestly i shouldn't even have touched this file, i just added the function to strip trailing slashes and added it here too because the fact you're not allowed to put a trailing slash on this particular env var has caused confusion in the past.

either way if the body isn't as expected the functions that call this endpoint (which is just a proxy to the multiplayer API) will fail parsing the response and that'll trigger error handling, so I think it's OK.

PR #1416, opened 2025-01-13
1361 add projects widget to filecoin explore page

mhgbrown:
This PR adds the recently claimed projects section to the explore page on Filecoin. It also DRYs up the logic to fetch projects and adds caching for the Filecoin explore page data.


src/routes/(pages)/app/(app)/components/filecoin-explore-page.svelte
```diff
@@ -1,3 +1,8 @@
+<script lang="ts" context="module">
+  // DEFAULT_EXPLORE_PAGE_FEATURED_PROJECT_FRAGMENT
+  // shared with load-default-explore-page
+</script>
```
efstajas, line 4: I think _ideally_ there would be a component for `ExplorePageFeaturedProjectSection` that would export this fragment, and that would simply be imported both here and on the default explore page. But I'm also fine with this for now 👍

efstajas (approved):
LGTM 🚀 let's quickly test on staging filecoin and merge if all good.

src/routes/(pages)/app/(app)/components/filecoin-explore-page.svelte
mhgbrown: Dang, of course that makes total sense. I'll do a fast follow

PR #1418, opened 2025-01-14
1361 add projects widget to filecoin explore page 2

mhgbrown:


src/styles/projects.css
```diff
@@ -0,0 +1,8 @@
+.projects-grid {
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(14rem, 1fr));
+  gap: 1rem;
+  max-width: 100%;
+  position: relative;
+  padding: 4px;
+}
```
efstajas, line 8: why the global style? don't see any reason this can't be on the new `RecentlyClaimedProjects` component

efstajas (changes requested):

src/styles/projects.css
mhgbrown: It's used by default-explore-page, filecoin-explore-page, and recently-claimed-projects

mhgbrown: I refactored project grids into their own component

efstajas (approved):

PR #1443, opened 2025-02-05
Jason/total dripped update

efstajas:
Update the total dripped approx with latest streams and gives.

To make this easier in the future, adds some additional data to stream page developer section, including full timeline payload: 

Screenshot 2025-02-05 at 15 24 17

Also adds:
- Ability to set an end date for a stream in the total dripped approx so that we don't have to manually make a change here to prevent overshooting of the approximation. Can be set to either the date the stream is expected to run out of funds, or a set end date, if any.
- Ability to use tokens that belong to a network other than the one the app is connected to for approximations.

...... We really need to start calculating this value dynamically at some point!


mhgbrown (approved):
Looks goody!

PR #1446, opened 2025-02-07
add metis support, configurable explore pages

efstajas:
MERGE AFTER #1428 

(Check expected to fail bc metis API not deployed yet on mainnet)

Changes the "Filecoin explore page" into a re-usable "Distribution explore page" that we can apply to networks that we deploy on for the RPGF distribution usecase. Adds plumbing for assigning different re-usable "explore page variants" to specific networks and configuring them in `/routes/(pages)/app/(app)/+page.server.ts`.

Also adds configuration for final Metis deployment.


src/lib/stores/wallet/network.ts
```diff
@@ -395,13 +396,63 @@ export const NETWORK_CONFIG: ValueForEachSupportedChain<Network> = {
...
+    label: 'Metis',
+    token: 'METIS',
+    id: '0x440',
+    rpcUrl: 'https://andromeda.metis.io/?owner=1088',
+    icon: Metis,
+    color: '#00D2FF',
+    isTestnet: false,
+    subdomain: 'metis.drips.network',
+    gqlName: SupportedChain.Metis,
+    autoUnwrapPairs: [{ name: 'Filecoin', nativeSymbol: 'FIL', wrappedSymbol: 'WFIL' }],
```
mhgbrown, line 411: Is this right for Metis?

src/lib/stores/wallet/network.ts
```diff
@@ -395,13 +396,63 @@ export const NETWORK_CONFIG: ValueForEachSupportedChain<Network> = {
...
+    displayNetworkPicker: true,
+    applyGasBuffers: false,
+    explorer: {
+      name: 'Blockscout',
+      linkTemplate: (txHash: string) => `https://filecoin.blockscout.com/tx/${txHash}`,
```
mhgbrown, line 416: Is this the right explorer `explorer.linkTemplate`?

src/routes/(pages)/app/(app)/+page.server.ts
```diff
@@ -1,20 +1,89 @@
...
+          '80928956680761126169933069488824219882230509123653596947248123478016', // websockets
+        ],
+      }),
+  },
+  80002: { variant: 'default', loadFn: loadDefaultExplorePageData },
+  11155420: { variant: 'default', loadFn: loadDefaultExplorePageData },
+  11155111: { variant: 'default', loadFn: loadDefaultExplorePageData },
+  31337: { variant: 'default', loadFn: loadDefaultExplorePageData },
+  84532: { variant: 'default', loadFn: loadDefaultExplorePageData },
+  314: {
```
mhgbrown, line 52: What do you think about constantizing this numbers so that we can see what networks they are directly?
`[CHAIN_ID_SEPOLIA]: { ...stuff }`

and getting crazier...

```js
import { Sepolia } from 'path/to/supported-networks'

[Sepolia.CHAIN_ID]: {...stuff}
```

Could be cute to blow them up into entities/classes

src/routes/(pages)/app/(app)/components/load-distribution-explore-page-data.ts
```diff
@@ -0,0 +1,43 @@
...
+    fetchProjectsQuery + featuredDripListQuery,
+    [Object.entries(projectsParameters)],
+    'explore-page',
+  );
+
+  const { featuredListId, welcomeCardConfig, showRecentProjects } = config;
+
+  const [blogPosts, projects, dripList] = await cached(redis, cacheKey, 6 * 60 * 60, async () =>
+    Promise.all([
+      fetchBlogPosts(),
```
mhgbrown, line 31: ❤️ fetchBlogPosts

src/lib/stores/wallet/network.ts
efstajas: no!! bad merge here from the old `metis` branch. Thanks for spotting, fixing 👍

src/routes/(pages)/app/(app)/+page.server.ts
efstajas: Yeah that'd probably be nice. Maybe we could just have an `enum Network` instead of the `SUPPORTED_CHAINS` array in `network.ts` and use that everywhere? 

Kinda feeling a separate task for that though, since we'd want to change it in quite a few places

mhgbrown: Love an enum and a separate task would be appropriate, for sure!

PR #1453, opened 2025-02-13
fix: #1444

### LINKED ISSUE #1444: Start date and end date missing in scheduled streams
A stream _scheduled_ for the future should also display the start date and end date.

Example: https://www.drips.network/app/131394345097299958177194341257132791559337576226/tokens/0xe29797910d413281d2821d5d9a989262c8121cc2/streams/3887785930

Image

efstajas:
Changes some of the values displayed on the stream page to make them show all relevant data in different situations:

- Adds dedicated "Created at" value that always shows when the stream was created on-chain
- Adds dedicated "Configured start date" and "Configured end date" values that appear if start- and end dates have been set explicitly
- Adds a "Runs / ran out of funds at" value that shows the date when the stream will run out of funds (if any)
- Adds a warning tooltip next to "Configured end date" if the end date is later than when the stream will run out of funds

Screenshot 2025-02-13 at 10 42 00

Resolves #1444


mhgbrown (approved):
:shipit:

PR #1455, opened 2025-02-13
e2e setup

efstajas:
Adds e2e test setup with playwright. Also disables stepper animations if `prefers-reduced-motion` is true.

See `DEVELOPMENT.md` for instructions.

For now only includes a test for claiming a project, we should add many more afterwards.

Includes github action that executes E2E tests. Right now the tests are executed against dev server, which is not ideal, but works and is unlikely to cause false _positives_, at least.

I'm creating follow-up tasks for 1) writing more tests, and 2) getting it to run against a prod image.


docker/update-repo-owner.sh
```diff
@@ -1,10 +1,51 @@
...
+  printf "Enter the owner address that should be set for the repo.\n"
+  read -p "- New owner address to set: " new_owner_address
+fi
 
 printf "\nSending the transaction. This will only work if you have the local env running (npm run dev:docker)...\n\n"
 
-exec docker run --rm --network host --entrypoint cast ghcr.io/foundry-rs/foundry:latest send --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9 "updateOwner(address,uint256,address)" 0x971e08fc533d2A5f228c7944E511611dA3B56B24 $repo_account_id $new_owner_address
+if [ "$use_cast" = true ]; then
+  printf "Using local cast command...\n\n"
+  cast send --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
```
mhgbrown, line 39: What's `cast`?

docs/DEVELOPMENT.md
```diff
@@ -53,8 +52,48 @@ To claim a project during local dev mode:
...
+
+### Writing tests
+
+E2E tests go into the `/tests/` directory.
+
+Tip: Run `npx playwright codegen http://localhost:5173` after starting the local dev stack to bring up a test recorder that automatically writes tests for you as you interact with the app.
+
+### Limitations
+
+Tests all run against the same local dev stack, which is created before tests are triggered, and destroyed afterwards. This means that separate tests need to be designed such that state changes they trigger on the local dev stack (e.g. projects being claimed, tokens being topped up, drip lists being created etc.) do not interfere with any other test.
```
mhgbrown, line 99: Good note!

tests/claim-project.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { test, expect } from '@playwright/test';
+import { execa } from 'execa';
+
+test('claim project flow', async ({ page }) => {
```
mhgbrown, line 4: Beauty

tests/claim-project.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  test.setTimeout(240_000);
```
mhgbrown, line 5: Were you going to look at production building to avoid this (assuming this is from dev optimization stuffs)?

tests/claim-project.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  await page.getByRole('textbox').first().press('Enter');
+  await page.getByRole('button', { name: 'Continue' }).click();
+  await page.getByRole('button', { name: 'Continue' }).click();
+  await page.getByRole('button', { name: 'Confirm in wallet' }).click();
+
+  await expect(page.getByTestId('current-tx')).toContainText('Finalizing verification', {
+    timeout: 60_000,
+  });
+
+  await execa`npm run dev:docker:update-repo-owner -- --accountId 80921553623925136102837120782793736893291544351678576578072673071408 --ownerAddress 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`;
```
mhgbrown, line 33: Any of those biggies worth constantizing to share?

mhgbrown (approved):
Looks fab! I left some comments and questions that aren't critical

docker/update-repo-owner.sh
efstajas: `cast` is a CLI part of `foundry`: https://book.getfoundry.sh/cast/ You can use it to send transactions. The command is usually just executed within the official foundry docker image so that you don't need to have it installed on your machine. I was struggling with getting it to work on CI, so tried an approach where the CI would install foundry and then run the command directly, but eventually I managed to get it to run it in Docker, so it wasn't necessary after all. Could remove this option, but doesn't hurt i guess

tests/claim-project.spec.ts
efstajas: That is actually necessary regardless, because our tests are necessarily somewhat slow running, just because we need to wait for transactions. This setTimeout defines the timeout for the _entire_ test, which is only 30 seconds by default, which is too little. It'd be necessary even if it ran against a prod image

tests/claim-project.spec.ts
efstajas: yeah... generally what's still missing is a way for the tests to manage addresses, so that you can test interactions between different users (e.g. user A sends B some funds, user B should be able to collect them etc.).

I'm going to create a follow-up to create a module for use in tests that can somehow make the app switch users, get the current address, etc., which would then be used here in place of the hardcoded addresses 👍

PR #1480, opened 2025-03-04
fix dev server not updating sk dev types

efstajas:
`.svelte-kit` dir, which includes dynamically created types for development (such as route params, page data types etc.) was being mounted in the dev container as an anonymous volume, causing it to not be written on the host. This means that types would never by updated during dev unless runing the dev server or svelte-kit sync on the host, outside of Docker.

This makes it so that the dev server running within Docker also updates the .svelte-kit dir on the host.

It also adds logic for always running the dev server as the same user that ran `npm run dev:docker` to fix permission issues.

Also adds an ASCII `GM` log 🌈

### DIFF
--- a/docker/dev-entrypoint.sh
+++ b/docker/dev-entrypoint.sh
@@ -12,10 +12,8 @@ printf "  \_____|_|  |_|\n\n"
                 
 echo "ℹ️ Starting with UID: $USER_ID, GID: $GROUP_ID"
 useradd -u $USER_ID -o -m drippy
-groupmod -g $GROUP_ID drippy
+groupmod -g $GROUP_ID drippy 2>/dev/null
 export HOME=/home/drippy
-
-ls -al | grep .svelte-kit
 
 /usr/sbin/gosu drippy npm run gql:build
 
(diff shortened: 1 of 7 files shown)


docker/dev-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+
+USER_ID=${LOCAL_UID:-9001}
+GROUP_ID=${LOCAL_GID:-9001}
+
+printf "\n   _____ __  __ \n"
+printf "  / ____|  \/  |\n"
+printf " | |  __| \  / |\n"
+printf " | | |_ | |\/| |\n"
+printf " | |__| | |  | |\n"
+printf "  \_____|_|  |_|\n\n"
```
mhgbrown, line 11: ❤️

docker/dev-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+                
+echo "ℹ️ Starting with UID: $USER_ID, GID: $GROUP_ID"
+useradd -u $USER_ID -o -m drippy
```
mhgbrown, line 14: 💀🫶

mhgbrown (approved):
LGTM

PR #1481, opened 2025-03-04
web3privacynow short URL

efstajas:
Adds a crude way of setting up short URLs for campaigns, before we move to a dynamic ENS-based approach.

Configures first redirect for web3privacynow campaign.

### DIFF
--- a/src/routes/(pages)/[path]/+page.ts
+++ b/src/routes/(pages)/[path]/+page.ts
@@ -1,0 +1,17 @@
+import { redirect } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+// Manually hardcoded short URL redirects for campaigns.
+// We want to move to a dynamic ENS-based solution long-term
+const SHORT_URLS: Record<string, string> = {
+  privacyproof24:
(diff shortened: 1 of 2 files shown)


src/routes/(pages)/[path]/+page.ts
```diff
@@ -0,0 +1,17 @@
...
+  privacyproof24:
+    'https://optimism.drips.network/app/drip-lists/46441013481627019632859175771245733399752255312769848791334977723541',
+};
+
+export const load: PageLoad = ({ params }) => {
+  if (params.path in SHORT_URLS) {
+    return redirect(301, SHORT_URLS[params.path]);
+  }
+
+  redirect(308, `/app/${params.path}`);
```
Comment (jtourkos) [inline src/routes/(pages)/[path]/+page.ts:16]: Not important, but for consistency I'd `return` here, also.

jtourkos (approved):

src/routes/(pages)/[path]/+page.ts
Comment (efstajas) [reply src/routes/(pages)/[path]/+page.ts:16]: yeah shouldn't matter. need to return above only so that the function stops. since this is the last statement it's ok 🤷

PR #1486, opened 2025-03-06
Clearly state settlement info when making donations

### LINKED ISSUE #1377: Clearly state the monthly settlement situation in the donation flows
Clearly state the monthly settlement situation in the donation flows (one-time and continuous).

we already have this pattern of those cards at the bottom on the review screen that say what happens after the transaction.

Should prob just add that to the donation flows

### For instant donations
**Project**
Image

**Drip List**
Image

### For continuous donations

**Project**
Not possible to continuously donate to a project

**Drip List**
Image

efstajas:
Component-izes the "whats next" sections across different flows and adds one to OTD and CD flows, clearly stating the settlement situation for the current network.

### Continuous donation to Drip List
image

### One-time donation to Drip List
image

### One-time donation to Project
image

Resolves #1377


src/lib/flows/claim-project-flow/steps/review/review.svelte
```diff
@@ -41,6 +40,9 @@
   import network from '$lib/stores/wallet/network';
   import { mapSplitsFromListEditorData } from '$lib/components/splits/utils';
   import sanitize from 'sanitize-html';
+  import WhatsNextSection from '$lib/components/whats-next/whats-next-section.svelte';
+  import WhatsNextCard from '$lib/components/whats-next/whats-next-card.svelte';
+  import WhatsNextItem from '$lib/components/whats-next/whats-next-item.svelte';
```
mhgbrown, line 45: Nice componentization!

src/lib/flows/create-donation/input-details.svelte
```diff
@@ -82,7 +96,7 @@
     createDonation(
       dispatch,
       recipientAccountId,
-      $context.selectedTokenAddress?.[0] ?? unreachable(),
+      selectedTokenAddress ?? unreachable(),
```
mhgbrown, line 99: Just want to point out that this is kind of similar to `shouldNeverHappen()` (or whatever it was called) in that it doesn't provide much help if it is indeed unreachable.

src/lib/flows/create-donation/input-details.svelte
```diff
@@ -115,9 +129,46 @@
     bind:amount
     bind:selectedTokenAllowance
   />
+
+  {#if amount && selectedToken}
```
mhgbrown, line 133: I know this isn't the case yet and a question for later but: if and how does this change for ecosystems? Is there anything special or is it basically like a drip list?

mhgbrown (approved):
Looking goody. Just some random comments and questions

src/lib/flows/create-donation/input-details.svelte
efstajas: true. it's somewhat different here in that we only use it on this repo if the value in question must be defined for this code to even be executed, which is what's the case here (createDonation called by confirm in wallet button which is disabled unless selectedTokenAddress is defined). But it's still a possible pitfall. Don't think it should be fixed in this PR but we could consider making the error message mandatory same way we're planning to on other repos

src/lib/flows/create-donation/input-details.svelte
efstajas: this should just work for ecosystems, since yeah, it's just an NFT Driver ID.

PR #1491, opened 2025-03-13
automatically refresh supporters section after OTD

efstajas:
Makes it so that one-time donations immediately appear on a drip list / project page after being sent, no need to manually refresh

### DIFF
--- a/src/lib/flows/create-donation/methods/create-donation.ts
+++ b/src/lib/flows/create-donation/methods/create-donation.ts
@@ -8,10 +8,78 @@ import type { OxString } from '$lib/utils/sdk/sdk-types';
 import { populateErc20WriteTx } from '$lib/utils/sdk/erc20/erc20';
 import { populateAddressDriverWriteTx } from '$lib/utils/sdk/address-driver/address-driver';
 import network from '$lib/stores/wallet/network';
+import expect from '$lib/utils/expect';
+import query from '$lib/graphql/dripsQL';
+import { gql } from 'graphql-request';
+import filterCurrentChainData from '$lib/utils/filter-current-chain-data';
+import { invalidateAll } from '$lib/stores/fetched-data-cache/invalidate';
+import type {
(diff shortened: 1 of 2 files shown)


src/lib/flows/create-donation/methods/create-donation.ts
```diff
@@ -8,10 +8,78 @@ import type { OxString } from '$lib/utils/sdk/sdk-types';
...
+          date
+        }
+      }
+    }
+  }
+`;
+
+function checkDonation(
+  ownAccountId: string,
+  supportAccountId: string,
```
jtourkos, line 72: Since we have it I'd type it as `AccountId` (which is a `string` ofc).

src/lib/flows/create-donation/methods/create-donation.ts
```diff
@@ -8,10 +8,78 @@ import type { OxString } from '$lib/utils/sdk/sdk-types';
...
+  supportDate: number,
+  blockTimestamp: number,
+) {
+  return ownAccountId === supportAccountId && supportDate / 1000 === blockTimestamp;
+}
 
 export default function (
   dispatch: ReturnType<typeof createEventDispatcher<StepComponentEvents>>,
   recipientAccountId: string,
+  recipientType: 'AddressDriverAccount' | 'Project' | 'NftDriverAccount',
```
jtourkos, line 82: Why not just `Address` since we are not using the "driver names"?

src/lib/flows/create-donation/methods/create-donation.ts
```diff
@@ -68,28 +137,69 @@ export default function (
...
+                30000,
+                1000,
+              );
+              break;
+            }
+          }
+
+          await invalidateAll();
+        } catch {
+          return;
```
jtourkos, line 201: Should we log something here?

jtourkos (approved):
LGTM! Left some minor suggestions.

PR #1492, opened 2025-03-14
Support ens resolution on chains other than mainnet

efstajas:
Adds safe resolution of ENS names (and ens name display of addresses) on chains other than Eth Mainnet.

```
Address -> Name
- just look up mainnet

Name -> Address
- also just lookup on mainnet, however:
- check if an explicit address for the chain in question is configured on the ENS record, use that if yes
- if no:
  - check if address it resolves to on mainnet is a contract 
    - if yes, don't resolve
    - if no, resolve to the address specified on mainnet, but display a warning about it
```


src/lib/stores/ens/ens.ts
```diff
@@ -0,0 +1,64 @@
...
+  mainnetProvider: AbstractProvider,
+  chainId: number,
+  name: string,
+) {
+  if (chainId === 1) {
+    return await mainnetProvider.resolveName(name);
+  }
+
+  const address = await mainnetProvider.resolveName(name);
+  if (!address) return undefined;
```
mhgbrown, line 40: Just out of curiosity, why do we returned undefined in one case and the straight return value in another?

mhgbrown (approved):
lgtm with a question

src/lib/stores/ens/ens.ts
efstajas: You're right this doesn't make any sense. Fixing Monday

Issue #1439, opened 2025-02-03
Introduce proper error handling within `transact-step` before


efstajas:
Currently, when an error occurs during a transact step's `before` function, it is uncaught and the UI keeps displaying the "Preparing" spinner. This means that if an error occurs in this step, 1) the user is not aware of the error, and 2) the UI is locked and just keeps looking as if it's still doing something. The user will think the process just takes a while and it wastes their time 👎 

To fix, refactor `stepper` to have generic error handling. It currently can already handle errors from `Await` steps by having the `AwaitResult` function return an error, and then letting the user retry via `handleAwaitErrorRetry`. Instead, there should be a generic `Error` event that can be passed from steps to the stepper, and is handled there by transitioning to the Error step. Both `Await` and `Transact` steps then must be refactored to emit those generic `Error` events on error. The `Error` step must be refactored to emit a generic `Retry` event when the user clicks "Try again".

PR #1501, opened 2025-03-27
fix: #1439

efstajas:
Triggers an `await-error-step` if the `before` fn on a `transact` step fails. This allows the user to see the error and retry. Before this, it would simply stay on this step loading forever without any indication of an error.


mhgbrown (approved):
Looks good. I haven't tried it out though

jtourkos (approved):

PR #1504, opened 2025-03-28
fix: #1380

### LINKED ISSUE #1380: Stop returning 404 for deleted GH repos
From @efstajas:

> i suppose if we were to stop returning 404 for those (claimed projects that had their repo deleted afterwards), the owners of those orphan projects could just hide them
like, render the actual project page but with a warning banner that says "the repo for this has been deleted on github", but the owner can still edit the project, including hiding it

efstajas:
resolves #1380

### DIFF
--- a/src/lib/components/stepper/components/transact-step.svelte
+++ b/src/lib/components/stepper/components/transact-step.svelte
@@ -94,9 +94,7 @@
   let error: Error | undefined;
   let transactionsTimeline: TransactionTimelineItem[] = [];
 
-  onMount(async () => {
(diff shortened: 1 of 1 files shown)


jtourkos (commented):
Should we somehow show that the project "doesn't exist" at the Drip List level as well?

efstajas:
@jtourkos it'll be hard bc we'd need to make a query to GH for every entry on a list you're viewing. so i think ideally yeah but we can't and it's not a deal-breaker.

PR #1536, opened 2025-05-06
disable redis during build

efstajas:
Disables redis entirely during build time, finally allowing for using private / internal networking on app instances. This should result in a nice performance boost.


jtourkos (approved):
👍

PR #1575, opened 2025-07-23
Update NFT Metadata Driver to v6

mhgbrown:
Ignoring the parsing of Ecosystem-related metadata for now because we can't edit them.

### DIFF
--- a/src/lib/utils/metadata/NftDriverMetadataManager.ts
+++ b/src/lib/utils/metadata/NftDriverMetadataManager.ts
@@ -19,7 +19,8 @@ export default class NftDriverMetadataManager extends MetadataManagerBase<
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/lib/utils/metadata/NftDriverMetadataManager.ts
```diff
@@ -17,6 +18,29 @@ export default class NftDriverMetadataManager extends MetadataManagerBase<
     super(nftDriverAccountMetadataParser, nftDriver);
   }
 
+  /**
+   * Ensures that ecosystem metadata is not fetched. We shouldn't be handling it at this point.
```
jtourkos, line 22: Minor suggestion: "fetches NFTDriver metadata but ensures that...." you get the idea.

jtourkos (approved):

PR #1598, opened 2025-08-20
1596 navigating between projects does not update share modal image

mhgbrown:
This PR also reverts changes to actions referencing an old PR environment and adds `ECOSYSTEM_` env vars for the action of `check.yml`.


efstajas (approved):
LGTM, before merging can you run headless tests (with EVENT_PROCESSOR_BRANCH=ecosystems) locally for good measure?

Issue #1526, opened 2025-04-24
Empty block content on top of page


nya-elimu:
Sometimes when opening a drip list page or project page, I have to scroll down before the content becomes visible:

![Image](https://github.com/user-attachments/assets/192e0a36-5993-47c2-b711-76fa31ab7d86)

Seems like the reason is an animation `div` that isn't being properly removed after the page has finished loading:

![Image](https://github.com/user-attachments/assets/da6ce34f-3d12-4ef0-bf04-2cc23b893d62)

efstajas:
Hey @nya-elimu, yeah, we've been battling this bug for a while. It's some very weird thing happening deep in SvelteKit somewhere. 

Luckily, a fix for it (switching to webview API) is included in #1520, which will merge sometime this week 🙌

efstajas:
It turns out this issue still appears in some scenarios, specifically when navigating between pages in rapid succession. Although now the old page is at least not invisible anymore.

I found this: https://stackoverflow.com/questions/70051025/sveltekit-adds-new-page-on-top-of-old-one

Re-opening to give it (avoiding DOM mutation during transition) a try.

PR #1599, opened 2025-08-21
fix: #1526

efstajas:


efstajas:
It seems that this fix isn't successful, I managed to recreate the bug again on staging. Investigating further.

efstajas:
that seems to have done it! no longer able to repro the issue on [staging](https://drips-mainnet-app-staging.up.railway.app/app/drip-lists)

PR #1605, opened 2025-08-26
1568 linked identities and orcids

mhgbrown:


brandonhaslegs:
It's not possible to create an ecosystem, we need to rewrite this 
image

src/routes/(pages)/app/(app)/[accountId]/+page.svelte
```diff
@@ -47,99 +51,142 @@
...
+        accountId={data.profileData.account.accountId}
+      />
+      <Balances
+        collapsable
+        collapsed={profileChainData.balances.length === 0}
+        userBalances={profileChainData.balances}
+        accountId={data.profileData.account.accountId}
+      />
+    </div>
+    <aside class="article-sidebar">
```
Comment (efstajas) [inline src/routes/(pages)/app/(app)/[accountId]/+page.svelte]: do i understand correctly that this would be shown even if orcid is `false` for the current network? shouldn't show this then.

src/lib/components/list-editor/list-editor.svelte
```diff
@@ -30,6 +32,7 @@
   export let allowDripLists = true;
   export let allowProjects = true;
   export let allowAddresses = true;
+  export let allowOrcids = true;
```
efstajas: maybe i'm missing something, but i don't see this being set to `false` in list editor usages? might be best to set to false by default for now to ensure no list-editors anywhere suddenly start accepting orcids, especially on networks where it's not yet enabled

PR #1618, opened 2025-09-09
optimize lp bundle + bundle analyzer

efstajas:
Adds a bundle analyzer build step that outputs a stats.html page that visualizes built bundles. 

Based on that I removed some of the HUGE svg components we had on the LP and replaced them with images, which should have a significant effect on our Lighthouse performance score.


mhgbrown (approved):
lgtm!

PR #1707, opened 2025-10-29
Fix 1639 use virtual scrolling in list-select, like in list-editor

mhgbrown:
Replaced standard rendering with virtual scrolling in the `list-select` component to improve performance with large item lists (10,000+), matching the implementation used in `list-editor`. There's also now a many items example for `list-select` in the component showcase.


efstajas:
@mhgbrown here's a screencap 

https://github.com/user-attachments/assets/b2c271f1-4e4f-4533-bef0-59e501e60c39

PR #1937, opened 2026-07-09
feat(wave): nest error pages under wave layout, fix 500 on non-existent user profile

jtourkos:
Implements the frontend part of drips-network/wave#665.

Backend counterpart: drips-network/wave#697

## What
- **Fix 500 on 404 user profile** The profile loader now fetches the user first and 404s early. Non-existent users return 404, so we must not surface points, issues, or org data from the sibling endpoints for them (previously those ran in a `Promise.all` and could reject first, yielding a 500 instead of a clean 404).
- **Error pages nested under the wave main layout.** Added a `(base-layout)` error boundary so errors thrown anywhere under the wave shell render *inside* the wave header/nav chrome, instead of bubbling to the bare app-wide error page.
- Added a shared `wave-error` component and refactored the near-identical existing `single-issue-error` page to reuse it.

## Verification
- `eslint` clean on all changed files.
- `svelte-check` clean on all changed files (repo-wide errors are pre-existing `$env/static/public` env-var issues, unrelated to this change).

### DIFF
--- a/src/routes/(pages)/wave/(base-layout)/users/[userId]/+page.ts
+++ b/src/routes/(pages)/wave/(base-layout)/users/[userId]/+page.ts
@@ -1,30 +1,47 @@
 import { getIssues } from '$lib/utils/wave/issues.js';
 import { getPointsBalanceForUser } from '$lib/utils/wave/points.js';
 import { getUser, getUserOrgs } from '$lib/utils/wave/users.js';
-import { error } from '@sveltejs/kit';
+import { error, redirect } from '@sveltejs/kit';
+import z from 'zod';
 
 export const load = async ({ params, fetch }) => {
   const { userId } = params;
 
-  // Fetch everything in parallel. When the backend won't expose a user it 404s
-  // the user-scoped endpoints, so we keep the orgs lookup from rejecting the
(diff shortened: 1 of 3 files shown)


src/routes/(pages)/wave/(base-layout)/users/[userId]/+page.ts
```diff
@@ -3,16 +3,23 @@ import { getPointsBalanceForUser } from '$lib/utils/wave/points.js';
...
+  const profileUserData = await getUser(fetch, userId);
+  if (!profileUserData) {
+    throw error(404, 'User not found');
+  }
+
+  const [pointsBalance, resolvedIssues, orgs] = await Promise.all([
     getPointsBalanceForUser(fetch, userId),
     getIssues(fetch, { limit: 50 }, { assignedToUser: userId, state: 'closed' }),
     getUserOrgs(fetch, userId),
   ]);
```
Comment (efstajas) [inline src/routes/(pages)/wave/(base-layout)/users/[userId]/+page.ts]: not a fan of reducing parallelization here - we should be able to fetch all in parallel still, but handle errors gracefully, and return 404 if getUser is null

efstajas (changes requested):

efstajas (approved):

PR #1955, opened 2026-08-28
fix(wave): show recovery copy when a login email is already linked

efstajas:
## What

When a GitHub account's primary email already belongs to a different Wave account, login cannot succeed — `users.email` is unique, so the same login fails the same way forever.

Until now the callback page rendered the generic branch:

> Something went wrong. Please try again, and if this problem persists, contact us at support@drips.network.

…behind a **Try again** button. For this failure that button can never work, so people loop on it and give up. Production shows 3-5 attempts in ~13 minutes before abandonment.

## Change

The backend now returns `409` with `code: "email_already_linked"` (drips-network/wave#790). This PR surfaces it:

- `EmailAlreadyLinkedError` in `call.ts`, matched on the machine-readable `code` rather than the prose — same pattern as the existing suspended / restricted / checkpoint errors.
- The backend's message (which names the address) is carried through, with a generic fallback if the body isn't the expected shape.
- `perform-login` rethrows it instead of flattening it into "GitHub may be experiencing issues".
- The callback page renders a `LargeEmptyState` with the two actions that can actually help — **Use a different account** and **Contact support** — instead of a dead "Try again".

## Tests

6 new unit tests in `call.unit.test.ts`: the 409 + code maps to the typed error, the backend message is surfaced, the fallback copy applies when the body isn't JSON, other 409s stay generic, the code on a non-409 status is ignored, and 403 suspensions still route to `AccountSuspendedError`.

`npm run check` reports no errors in the touched files.

## Deploy order

Merge drips-network/wave#790 first. Safe either way — without the backend change this code path simply never triggers and today's generic message still shows.

### DIFF
--- a/src/lib/utils/wave/call.ts
+++ b/src/lib/utils/wave/call.ts
@@ -23,6 +23,23 @@ export class AccountRestrictedError extends Error {
 }
 
 /**
+ * Thrown for the 409 the API returns when the GitHub account's primary email is
+ * already linked to a *different* Wave account.
+ *
+ * Terminal by nature: `users.email` is unique, so the same login will fail the
+ * same way forever. Callers must surface recovery copy rather than the generic
+ * "something went wrong, try again" — a retry can never succeed.
+ */
+export class EmailAlreadyLinkedError extends Error {
+  constructor(message?: string) {
+    super(
+      message ?? 'The email on this GitHub account is already linked to a different Wave account.',
+    );
+    this.name = 'EmailAlreadyLinkedError';
+  }
+}
+
+/**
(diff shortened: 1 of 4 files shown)


efstajas:
Closing — the backend is no longer sending anything for this to render.

drips-network/wave#790 now fails the email-collision login with the same opaque message as any other login failure, deliberately. We don't gate on GitHub's `verified` flag, and a production check found 7 of 9431 users with an unverified primary email, so a distinguishable response would have been an account-existence oracle for any address an attacker cared to name.

The generic copy already on the callback page is the intended behaviour. Affected users are pointed at support, who resolve it from a new structured log line.

